PR #7, opened 2017-05-06
Basic Peer

bfbachmann:
# What Changed:
- Added `cumuluspeer` with `MakeBasicHost` functionality 
- Updated `main.go` to take host port number, IP address and target peer to connect to as command line arguments
- Added tests for `cumuluspeer`
- Added documentation for `cumuluspeer` to Wiki

# Remaining TODOs
- Update main and `BasicStreamHandler` to actually communicate useful information between peer and host


cumuluspeer/peer.go
```diff
@@ -0,0 +1,145 @@
+package cumuluspeer
```
jordanschalm: Just call this `peer`. We know it's cumulus because inside the cumulus folder. Short and succinct package names

cumuluspeer/peer.go
```diff
@@ -0,0 +1,145 @@
...
+	pstore "github.com/libp2p/go-libp2p-peerstore"
+	swarm "github.com/libp2p/go-libp2p-swarm"
+	ma "github.com/multiformats/go-multiaddr"
+	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+)
+
+const (
+	DefaultPort 	= 8765 // Cumulus peers communicate over this TCP port
+	CumulusProtocol	= "/cumulate/0.0.1" // Cumulus communication protocol
+	DefaultIP		= "127.0.0.1" // Default Host IP address if none given
```
jordanschalm: Run gofmt here.

cumuluspeer/peer.go
```diff
@@ -0,0 +1,145 @@
...
+	net "github.com/libp2p/go-libp2p-net"
+	pstore "github.com/libp2p/go-libp2p-peerstore"
+	swarm "github.com/libp2p/go-libp2p-swarm"
+	ma "github.com/multiformats/go-multiaddr"
+	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+)
+
+const (
+	DefaultPort 	= 8765 // Cumulus peers communicate over this TCP port
+	CumulusProtocol	= "/cumulate/0.0.1" // Cumulus communication protocol
```
jordanschalm: Why `cumulate` instead of `cumulus`?

cumuluspeer/peer.go
```diff
@@ -0,0 +1,145 @@
...
+		return
+	}
+
+	log.Info("Done now. Bye!")
+}
+
+// Create a Cumulus host.
+// This may throw an error if we fail to create a key pair, a pid, or a new
+// multiaddress.
+func MakeBasicHost(ip string, port int) (host.Host, pstore.Peerstore, error) {
```
jordanschalm: You should wrap `host.Host` with your own struct `peer.Peer` and return that here. Also use the pattern `NewPeer` rather than `MakePeer`

cumuluspeer/peer.go
```diff
@@ -0,0 +1,145 @@
...
+func BasicStreamHandler(s net.Stream) {
+	log.Info("Setting basic stream handler.")
+	defer s.Close()
+	doCumulate(s)
+}
+
+// Communicate with peers.
+// TODO: Update this to do something useful. For now it just reads from the
+// stream and writes back what it read.
+func doCumulate(s net.Stream) {
```
jordanschalm: This should be a method on the `Peer` struct
`func (p *Peer) Receive(s net.String) {...}`

cumuluspeer/peer.go
```diff
@@ -0,0 +1,145 @@
...
+	buf := bufio.NewReader(s)
+	str, err := buf.ReadString('\n')
+	if err != nil {
+		log.Error(err)
+		return
+	}
+
+	log.Info("Read: %s", str)
```
jordanschalm: We should start making a distinction between `Log.Debug/Error/Warn`. This sort of stuff should be `Debug` or omitted

cumuluspeer/peer.go
```diff
@@ -0,0 +1,145 @@
...
+	_, err = s.Write([]byte(str))
+	if err != nil {
+		log.Error(err)
+		return
+	}
+
+	log.Info("Done now. Bye!")
```
jordanschalm: Debug here as well

cumuluspeer/peer.go
```diff
@@ -0,0 +1,145 @@
...
+	fullAddr := addr.Encapsulate(hostAddr)
+	log.Info("I am ", fullAddr)
+
+	// Add this host's address to its peerstore (avoid's net/identi error)
+	ps.AddAddr(pid, fullAddr, pstore.PermanentAddrTTL)
+
+	return basicHost, ps, nil
+}
+
+// Extracts target's the peer ID and multiaddress from the given multiaddress.
```
jordanschalm: ExtractPeerInfo targets...

cumuluspeer/peer_test.go
```diff
@@ -0,0 +1,38 @@
+package cumuluspeer_test
```
jordanschalm: all files in the same folder have the same package name

cumuluspeer/peer_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+    "testing"
+
+    cumuluspeer "github.com/ubclaunchpad/cumulus/cumuluspeer"
```
jordanschalm: So we don't need this import

cumuluspeer/peer_test.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// Tests if we can make a basic host on a valid TCP port
+func TestMakeBasicHostValidPort(t *testing.T) {
```
jordanschalm: TestNew...

main.go
```diff
@@ -0,0 +1,76 @@
...
+
+    // Set the host StreamHandler for the Cumulus Protocol and use
+    // BasicStreamHandler as its StreamHandler.
+    host.SetStreamHandler(cumuluspeer.CumulusProtocol,
+        cumuluspeer.BasicStreamHandler)
+
+    if *targetPeer == "" {
+        // No target was specified, wait for incoming connections
+        log.Info("No target provided. Listening for incoming connections...")
+        select {} // Hang until someone connects to us
```
jordanschalm: This will hang this goroutine forever, is that what we want?

main.go
```diff
@@ -0,0 +1,76 @@
...
+
+func main() {
+    log.Info("Starting Cumulus Peer")
+
+    // Get and parse command line arguments
+    // targetPeer is a Multiaddr representing the target peer to connect to
+    // when joining the Cumulus Network.
+    // port is the port to communicate over (defaults to peer.DefaultPort).
+    // ip is the public IP address of the this host.
+    targetPeer := flag.String("t", "", "target peer to connect to")
```
jordanschalm: If we know the IP and the port of the multiaddr can we infer the rest? A multiaddr as a CLI flag is an awkward ui

jordanschalm (changes requested):
Good job for your first Go code!

cumuluspeer/peer.go
```diff
@@ -0,0 +1,145 @@
...
+// This is called when a remote peer opens a new stream with the host that
+// SetStreamHandler() is called on.
+// We may want to implement another type of StreamHandler in the future.
+func BasicStreamHandler(s net.Stream) {
+	log.Info("Setting basic stream handler.")
+	defer s.Close()
+	doCumulate(s)
+}
+
+// Communicate with peers.
```
chadlagore: Re: https://golang.org/doc/effective_go.html#commentary. Function comments start with function name, ie `// doCumulate communicates with peers.` 

> The first sentence should be a one-sentence summary that starts with the name being declared. If every doc comment begins with the name of the item it describes, the output of godoc can usefully be run through `grep`.

peer/peer.go
```diff
@@ -0,0 +1,179 @@
...
+	p.Peerstore().AddAddr(peerid, targetAddr, pstore.PermanentAddrTTL)
+
+	log.Debug("Connected to Cumulus Peer:")
+	log.Debug("Peer ID:", peerid.Pretty())
+	log.Debug("Peer Address:", targetAddr)
+
+	// Open a stream with the peer
+	stream, err := p.NewStream(context.Background(), peerid,
+		CumulusProtocol)
+	if err != nil {
```
jordanschalm: Just `return stream, err` if you aren't handling error

peer/peer.go
```diff
@@ -0,0 +1,179 @@
...
+
+// Peer is a cumulus Peer composed of a host
+type Peer struct {
+	host.Host
+}
+
+// NewPeer creates a Cumulus host with the given IP addr and TCP port.
+// This may throw an error if we fail to create a key pair, a pid, or a new
+// multiaddress.
+func NewPeer(ip string, port int) (*Peer, error) {
```
jordanschalm: If you're making a package with one primary export (ie `Peer`) then just call the constructor `New`. Then you're calling `peer.New(...)` -- makes more sense semantically

PR #23, opened 2017-05-15
Better testing for `blockchain`

jordanschalm:


blockchain/test_utils.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func newTxHashPointer() TxHashPointer {
+	return TxHashPointer{
+		BlockNumber: mrand.Uint32(),
+		Hash:        newHash(),
+	}
+}
+
+func newTxOutput() TxOutput {
```
chadlagore, line 28: These helpers are great, but perhaps we could expose a parameter generate this `TxOutput` fixture in a way that is not random? Im thinking about testing valid transaction inputs and outputs 🤔

chadlagore (commented):
Looks great to me. But I'm a bit concerned about the validity of creating randomized test fixtures, makes replication of errors non-deterministic. 

It would be nice to have the option to pass parameters to some of these fixtures as well (validating inputs and outputs).

README.md
```diff
@@ -12,25 +12,21 @@ Install dependencies. We need to manually use version 0.11.1 of Glide temporaril
...
 ```sh
-cd $GOPATH/src/github.com/ubclaunchpad/cumulus
-glide install
+make
 ```
 
 ## Testing
 
 ```
-go test ./...
```
bfbachmann, line 35: Thanks!

blockchain/block.go
```diff
@@ -2,56 +2,71 @@ package blockchain
...
+)
+
 // BlockHeader contains metadata about a block
 type BlockHeader struct {
-	blockNumber uint32
-	lastBlock   Hash
-	miner       Wallet
+	BlockNumber uint32
+	LastBlock   Hash
+	Miner       Address
```
bfbachmann, line 22: What kind of address will this be?

blockchain/block.go
```diff
@@ -2,56 +2,71 @@ package blockchain
...
 func (b *Block) Encode(w io.Writer) {
-	gob.NewEncoder(w).Encode(b)
-}
-
-// Decode reads the marshalled block from the given io.Reader
-func (b *Block) Decode(r io.Reader) {
-	gob.NewDecoder(r).Decode(b)
+	err := gob.NewEncoder(w).Encode(b)
+	if err != nil {
+		fmt.Println(err.Error())
```
bfbachmann, line 63: We should probably change this to a `Log` statement at some point.

blockchain/test_utils.go
jordanschalm: Yeah good point

blockchain/blockchain.go
```diff
@@ -5,23 +5,40 @@ import (
...
 // Encode writes the marshalled blockchain to the given io.Writer
 func (bc *BlockChain) Encode(w io.Writer) {
 	gob.NewEncoder(w).Encode(bc)
 }
 
-// Decode reads the marshalled blockchain from the given io.Reader
-func (bc *BlockChain) Decode(r io.Reader) {
-	gob.NewDecoder(r).Decode(bc)
+// DecodeBlockChain reads the marshalled blockchain from the given io.Reader
+func DecodeBlockChain(r io.Reader) *BlockChain {
```
bfbachmann, line 38: Will we have to implement our own Blockchain io Reader to handle the format of the Blockchain in byte array format?

blockchain/hash.go
```diff
@@ -0,0 +1,28 @@
...
+func (h Hash) Marshal() []byte {
+	buf := make([]byte, HashLen)
+	for i, b := range h {
+		buf[i] = b
+	}
+	return buf
+}
+
+// Marshaller is any type that can convert itself to a byte slice
+type Marshaller interface {
```
bfbachmann, line 21: Love it!

blockchain/test_utils.go
```diff
@@ -0,0 +1,90 @@
+package blockchain
+
+import (
+	"crypto/ecdsa"
+	crand "crypto/rand"
+	"crypto/sha256"
+	mrand "math/rand"
+)
+
+func newHash() Hash {
```
bfbachmann, line 10: Does't one of the imports have a function that can do this for us?

blockchain/transaction.go
```diff
@@ -1,52 +1,60 @@
...
-	sender  Wallet
-	input   TxHashPointer
-	outputs []TxOutput
+	Sender  Address
+	Input   TxHashPointer
+	Outputs []TxOutput
+}
+
+// Len returns the length of a transaction body
+func (tb TxBody) Len() int {
```
bfbachmann, line 48: Perhaps `func (tb *TxBody) Len() int` would be better here? Not sure about the semantics, but something to think about.

bfbachmann: This goes for other exported functions on structs as well.

bfbachmann (approved):
Nice work my guy!

PR #28, opened 2017-05-17
ValidTransaction and ValidBlock

chadlagore:
**Status**: Ready to merge.

# What Changed

+ Implementing `ValidTransaction` and `ValidBlock`.

~Opening this up for visibility early! :relaxed: Not done yet.~

**I've changed the base branch to `dev`**

## TODO

- [x] Write tests.
- [x] `Signature` verification.


blockchain/blockchain.go
```diff
@@ -41,22 +42,66 @@ func DecodeBlockChain(r io.Reader) *BlockChain {
 	return &bc
 }
 
-// ValidTransaction checks whether a transaction is valid, assuming the blockchain is valid.
+// ValidTransaction checks whether a transaction is valid, assuming the
+// blockchain is valid.
 func (bc *BlockChain) ValidTransaction(t *Transaction) bool {
+
 	// Find the transaction input (I) in the chain (by hash)
+	var I *Transaction
```
jordanschalm: Maybe use `input` rather than `I`

blockchain/blockchain.go
```diff
@@ -41,22 +42,66 @@ func DecodeBlockChain(r io.Reader) *BlockChain {
...
+		}
+	}
+	var outAmount uint64
+	for _, output := range t.Outputs {
+		outAmount += output.Amount
+	}
+	if inAmount != outAmount {
+		return false
+	}
+
```
jordanschalm, line 75: We also need to check that input transaction hasn't been spent between when it was created and now, right?

blockchain/blockchain.go
```diff
@@ -41,22 +42,66 @@ func DecodeBlockChain(r io.Reader) *BlockChain {
...
 func (bc *BlockChain) ValidBlock(b *Block) bool {
+	// Check that block number is one greater than last block
+	lastBlock := bc.Blocks[b.BlockNumber-1]
+	if lastBlock.BlockNumber != b.BlockNumber-1 {
+		return false
+	}
+
+	// Verify every Transaction in the block.
 	for _, t := range b.Transactions {
 		if !bc.ValidTransaction(t) {
```
jordanschalm, line 96: What if two transactions in `b.Transactions` reference the same input?

blockchain/blockchain.go
```diff
@@ -41,22 +42,66 @@ func DecodeBlockChain(r io.Reader) *BlockChain {
...
 // ValidBlock checks whether a block is valid
 func (bc *BlockChain) ValidBlock(b *Block) bool {
+	// Check that block number is one greater than last block
+	lastBlock := bc.Blocks[b.BlockNumber-1]
+	if lastBlock.BlockNumber != b.BlockNumber-1 {
+		return false
+	}
+
+	// Verify every Transaction in the block.
 	for _, t := range b.Transactions {
```
jordanschalm, line 95: We also need to discuss how to spend coins that were mined, since there is no transaction for them to refer to. Do we have the miner insert a special transaction for their mined coins, or modify the `Transaction` struct to be able to refer to block rewards as inputs?

jordanschalm (approved):
Looks good. Couple extra things to check in validation (maybe add a test case for them) and we should add an issue for figuring out how to handle spending mined coins.

blockchain/blockchain.go
chadlagore: I feel like setting the `Transaction.Input` to some special value is a good way to go. Perhaps then we could get away with treating it as a normal block during verification (ie. I'd like to avoid special cases in the `VerifyX` functions). By convention, we could put the block reward in `Transactions[0]`.

blockchain/blockchain.go
chadlagore: 🥇 I've added a test case.

blockchain/blockchain.go
chadlagore: 🥇 I've added a test case.

jordanschalm:
Going to merge this now @chadlagore

blockchain/blockchain.go
jordanschalm: Added an issue for this

PR #34, opened 2017-05-20
[WIP] P2P rudimentary scaling

bfbachmann:
# What changed:
- Added `message` package
- Added `subnet` package
- Added `errors` package
- Refactor and add some functionality to `Peer` package

### DIFF
--- a/message/message.go
+++ b/message/message.go
@@ -1,73 +1,84 @@
 package message
 
 import (
-	"encoding/json"
-	"errors"
-	"strings"
+	"encoding/gob"
+	"io"
 )
 
-// Message types
-// NOTE: because of the way iota works, changing the order in which the
-// following constants appear will change their values, which may affect the
-// ability of your peer to communicate with others.
+type (
+	// Type specifies the type of a message.
+	Type int
+	// ResourceType specifies the type of a resource in a message.
+	ResourceType int
+)
+
 const (
-	// Send the multiaddress of a peer to another peer
-	PeerInfo = iota
-	// Request addressess of peers in the remote peer's subnet
-	RequestPeerInfo = iota
-	// Send information about a block that was just hashed
-	NewBlock = iota
-	// Request chunk of the blockchain from peer
-	RequestChunk = iota
-	// Advertise that we have a chunk of the blockchain
-	AdvertiseChunk = iota
-	// Send information about a new transaction to another peer
-	Transaction = iota
+	// MessageRequest messages ask a peer for a resource.
+	MessageRequest Type = iota
+	// MessageResponse messages repond to a request message with an error or a resource.
+	MessageResponse
+	// MessagePush messages proactively send a resource to a peer.
+	MessagePush
 )
 
-// Message is a container for information and its type that is
-// sent between Cumulus peers.
+const (
+	// ResourcePeerInfo resources contain a list of peers.
+	ResourcePeerInfo ResourceType = iota
+	// ResourceBlock resources contain a block in the blockchain.
+	ResourceBlock
+	// ResourceTransaction resources contain a transaction to add to the blockchain.
+	ResourceTransaction
+)
+
+// Message is a container for messages, containing a type and either a Request,
+// Response, or Push in the payload.
 type Message struct {
-	msgType int
-	content []byte
+	Type    Type
+	Payload interface{}
 }
 
-// New returns a pointer to a message initialized with a byte array
-// of content and a message type, or an error if the type is not one
-// of those defined above.
-func New(c []byte, t int) (*Message, error) {
-	switch t {
-	case PeerInfo:
-	case NewBlock:
-	case RequestChunk:
-	case AdvertiseChunk:
-	case Transaction:
-		break
-	default:
-		return nil, errors.New("Invalid message type")
+// New returns a new Message.
+func New(t Type, payload interface{}) *Message {
+	return &Message{
+		Type:    t,
+		Payload: payload,
 	}
-
-	m := &Message{msgType: t, content: c}
-	return m, nil
 }
 
-// Bytes returns JSON representation of message as a byte array, or error if
-// message cannot be marshalled.
-func (m *Message) Bytes() ([]byte, error) {
-	return json.Marshal(m)
+// Request is a container for a request payload, containing a unique request ID,
+// the resource type we are requesting, and a Params field specifying request
+// parameters. PeerInfo requests should send all info of all peers. Block requests
+// should specify block number in parameters.
+type Request struct {
+	ID           string
+	ResourceType ResourceType
+	Params       map[string]interface{}
 }
 
-// FromString parses a message in the form of a string and returns a pointer
-// to a new Message struct made from the contents of the string. Returns error
-// if string is malformed.
-func FromString(s string) (*Message, error) {
-	var msg Message
-	s = strings.TrimSpace(s)
-	err := json.Unmarshal([]byte(s), &msg)
-	return &msg, err
+// Response is a container for a response payload, containing the unique request
+// ID of the request prompting it, an Error (if one occurred), and the requested
+// resource (if no error occurred).
+type Response struct {
+	ID       string
+	Error    error
+	Resource interface{}
 }
 
-// Type returns msgType for message
-func (m *Message) Type() int {
-	return m.msgType
+// Push is a container for a push payload, containing a resource proactively sent
+// to us by another peer.
+type Push struct {
+	ResourceType ResourceType
+	Resource     interface{}
+}
+
+// Write encodes and writes the Message into the given Writer.
+func (m *Message) Write(w io.Writer) error {
+	return gob.NewEncoder(w).Encode(m)
+}
+
+// Read decodes a message from a Reader and returns it.
+func Read(r io.Reader) (*Message, error) {
+	var m Message
+	err := gob.NewDecoder(r).Decode(&m)
+	return &m, err
 }
--- a/message/message.go
+++ b/message/message.go
@@ -3,6 +3,9 @@ package message
 import (
 	"encoding/gob"
 	"io"
+
+	"github.com/google/uuid"
+	ma "github.com/multiformats/go-multiaddr"
 )
 
 type (
@@ -55,6 +58,17 @@ type Request struct {
 	Params       map[string]interface{}
 }
 
+// NewRequestMessage returns a new Message continaing a request with the given
+// parameters.
+func NewRequestMessage(rt ResourceType, p map[string]interface{}) *Message {
+	req := Request{
+		ID:           uuid.New().String(),
+		ResourceType: rt,
+		Params:       p,
+	}
+	return New(MessageRequest, req)
+}
+
 // Response is a container for a response payload, containing the unique request
 // ID of the request prompting it, an Error (if one occurred), and the requested
 // resource (if no error occurred).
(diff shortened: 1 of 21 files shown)


message/message.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Message types
+// NOTE: because of the way iota works, changing the order in which the
+// following constants appear will change their values, which may affect the
+// ability of your peer to communicate with others.
+const (
+	// Send the multiaddress of a peer to another peer
+	PeerInfo = iota
+	// Request addressess of peers in the remote peer's subnet
+	RequestPeerInfo = iota
```
jordanschalm: What is the difference between these two? When would we ever need only one peer's info

message/message.go
```diff
@@ -0,0 +1,73 @@
...
+	"strings"
+)
+
+// Message types
+// NOTE: because of the way iota works, changing the order in which the
+// following constants appear will change their values, which may affect the
+// ability of your peer to communicate with others.
+const (
+	// Send the multiaddress of a peer to another peer
+	PeerInfo = iota
```
jordanschalm: You only need to write iota the first time
Like:
```
const (
  a = iota
  b
  c
)
// a = 0, b = 1, c = 2
```

message/message.go
```diff
@@ -0,0 +1,73 @@
...
+	// Request addressess of peers in the remote peer's subnet
+	RequestPeerInfo = iota
+	// Send information about a block that was just hashed
+	NewBlock = iota
+	// Request chunk of the blockchain from peer
+	RequestChunk = iota
+	// Advertise that we have a chunk of the blockchain
+	AdvertiseChunk = iota
```
jordanschalm: I think we should assume that every node has the whole blockchain. That should be the first thing they do.

message/message.go
```diff
@@ -0,0 +1,73 @@
...
+// Bytes returns JSON representation of message as a byte array, or error if
+// message cannot be marshalled.
+func (m *Message) Bytes() ([]byte, error) {
+	return json.Marshal(m)
+}
+
+// FromString parses a message in the form of a string and returns a pointer
+// to a new Message struct made from the contents of the string. Returns error
+// if string is malformed.
+func FromString(s string) (*Message, error) {
```
jordanschalm: Should this be `FromBytes`/`Marshal` instead? Why is it a string?

peer/peer.go
```diff
@@ -89,38 +93,99 @@ func New(ip string, port int) (*Peer, error) {
...
+
+// Broadcast sends message to all peers this peer is currently connected to
+func (p *Peer) Broadcast(m msg.Message) error {
+	return errors.New("Function not implemented")
 }
 
 // ExtractPeerInfo extracts the peer ID and multiaddress from the
 // given multiaddress.
 // Returns peer ID (esentially 46 character hash created by the peer)
 // and the peer's multiaddress in the form /ip4/<peer IP>/tcp/<CumulusPort>.
```
bfbachmann, line 299: I'm not really sure that this function is necessary. I feel like it's doing much more work than it has to.

message/message.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, errors.New("Invalid message type")
+	}
+
+	m := &Message{msgType: t, content: c}
+	return m, nil
+}
+
+// Bytes returns JSON representation of message as a byte array, or error if
+// message cannot be marshalled.
+func (m *Message) Bytes() ([]byte, error) {
```
jordanschalm: `Marshal` for consistency

jordanschalm: Use `gob` for en/decoding

peer/peer.go
```diff
@@ -89,38 +93,99 @@ func New(ip string, port int) (*Peer, error) {
...
-// This should be passed as the second argument to SetStreamHandler().
-// We may want to implement another type of StreamHandler in the future.
+// opens a new stream with this peer.
+// This should be passed as the second argument to SetStreamHandler() after this
+// peer is initialized.
 func (p *Peer) Receive(s net.Stream) {
-	log.Debug("Setting basic stream handler.")
-	p.doCumulus(s)
-}
+	defer s.Close()
```
jordanschalm: Do you want to close the stream every time you receive? Is it a new stream each time?

peer/peer.go
```diff
@@ -89,38 +93,99 @@ func New(ip string, port int) (*Peer, error) {
...
+
+	// Get remote peer's full multiaddress
+	remoteMA, err := makeMultiaddr(
+		s.Conn().RemoteMultiaddr(), s.Conn().RemotePeer())
+	if err != nil {
+		log.Fatal("Failed to obtain valid remote peer multiaddress")
+	}
+
+	// Add the remote peer to this peer's subnet
+	err = p.subnet.AddPeer(remoteMA, s)
```
jordanschalm: What if the peer is already in our subnet? I feel like `Receive` should function like:
* unmarshal message
* switch on message type and call a subroutine

peer/peer.go
```diff
@@ -89,38 +93,99 @@ func New(ip string, port int) (*Peer, error) {
...
+	}
+	defer p.subnet.RemovePeer(remoteMA)
 
-// Communicate with peers.
-// TODO: Update this to do something useful. For now it just reads from the
-// stream and writes back what it read.
-func (p *Peer) doCumulus(s net.Stream) {
 	buf := bufio.NewReader(s)
-	str, err := buf.ReadString('\n')
+	strMsg, err := buf.ReadString('\n') // TODO: set timeout here
```
jordanschalm: Just get the bytes from the buffer here and add `unmarshal from byte slice` in the message object

peer/peer.go
```diff
@@ -89,38 +93,99 @@ func New(ip string, port int) (*Peer, error) {
...
 	if err != nil {
 		log.Error(err)
```
jordanschalm: You can do `log.WithError(err).Error("Error doing whatever")

peer/peer.go
```diff
@@ -14,23 +14,25 @@ import (
...
 	CumulusProtocol = "/cumulus/0.0.1"
-
 	// DefaultIP is the IP address new hosts will use if none if provided
 	DefaultIP = "127.0.0.1"
 )
 
 // Peer is a cumulus Peer composed of a host
 type Peer struct {
 	host.Host
+	subnet sn.Subnet
```
chadlagore: I _thiiiiink_ we're using capital letters, see https://github.com/ubclaunchpad/cumulus/pull/23/files `Subnet`

peer/peer.go
```diff
@@ -150,25 +215,48 @@ func ExtractPeerInfo(peerma string) (lpeer.ID, ma.Multiaddr, error) {
...
-// Connect adds the given multiaddress to p's Peerstore and opens a stream
-// with the peer at that multiaddress if the multiaddress is valid, otherwise
-// returns error.
-func (p *Peer) Connect(peerma string) (net.Stream, error) {
-	peerid, targetAddr, err := ExtractPeerInfo(peerma)
-	if err != nil {
-		return nil, err
+// advertisePeers writes messages into the given stream advertising the
+// multiaddress of each peer in this peer's subnet.
+func (p *Peer) advertisePeers(s net.Stream) {
```
jordanschalm: Why do we need to advertise peers? Why not just give our list of peers during connection, or accept requests for our list of peers. Seems like an an unnecessary complication

peer/peer.go
```diff
@@ -150,25 +215,48 @@ func ExtractPeerInfo(peerma string) (lpeer.ID, ma.Multiaddr, error) {
...
+	mAddrs := p.subnet.Multiaddrs()
+	log.Debug("Peers on this subnet: ")
+	for mAddr := range mAddrs {
```
jordanschalm: This seems like it should be refactored into `Broadcast`

peer/peer.go
```diff
@@ -150,25 +215,48 @@ func ExtractPeerInfo(peerma string) (lpeer.ID, ma.Multiaddr, error) {
...
+		mAddrString := string(mAddr)
+		log.Debug("\t", mAddrString)
+		message, err := msg.New([]byte(mAddrString), msg.PeerInfo)
+		if err != nil {
+			log.Error("Failed to create message")
```
jordanschalm: Same as above. Use `log.WithError(....)` Also two instances below

peer/peer.go
```diff
@@ -150,25 +215,48 @@ func ExtractPeerInfo(peerma string) (lpeer.ID, ma.Multiaddr, error) {
...
-	log.Debug("Peer ID:", peerid.Pretty())
-	log.Debug("Peer Address:", targetAddr)
-
-	// Open a stream with the peer
-	stream, err := p.NewStream(context.Background(), peerid,
-		CumulusProtocol)
+// makeMultiaddr creates a Multiaddress from the given Multiaddress (of the form
+// /ip4/<ip address>/tcp/<TCP port>) and the peer id (a hash) and turn them
+// into one Multiaddress. Will return error if Multiaddress is invalid.
+func makeMultiaddr(iAddr ma.Multiaddr, pid lpeer.ID) (ma.Multiaddr, error) {
```
jordanschalm: Use `NewMultiAddr` for consistency

peer/peer.go
```diff
@@ -150,25 +215,48 @@ func ExtractPeerInfo(peerma string) (lpeer.ID, ma.Multiaddr, error) {
...
+	strMA := fmt.Sprintf("%s/ipfs/%s", strAddr, strID)
+	mAddr, err := ma.NewMultiaddr(strMA)
+	return mAddr, err
+}
 
-	return stream, err
+func (p *Peer) handleMessage(m msg.Message, s net.Stream) {
+	switch m.Type() {
+	case msg.RequestPeerInfo:
+		p.advertisePeers(s)
```
jordanschalm: Why advertise to everyone when only one peer requests?

subnet/subnet.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	return nil
+}
+
+// RemovePeer removes the mapping with the key mAddr from the subnet if it
+// exists.
+func (sn *Subnet) RemovePeer(mAddr ma.Multiaddr) {
+	log.Debugf("Removing peer %s from subnet", mAddr.String())
+	delete(sn.peers, mAddr)
+	sn.numPeers--
```
jordanschalm: Don't separately keep track of `numPeers`, just use `len(sn.peers)`. It's O(1)

subnet/subnet.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// isFull returns true if the number of peers in the sunbet is at or over the
+// limit set for that subnet, otherwise returns false.
+func (sn *Subnet) isFull() bool {
```
jordanschalm: `Full` not `isFull`

jordanschalm (changes requested):
Highest Quality 👌

chadlagore (approved):

message/message.go
bfbachmann: Yea I think this should be `Unmarshal` and should take a `[]byte` as an argument.

peer/peer.go
bfbachmann: This function will only terminate when we are fully finished communicating with the peer on the other end of the stream. That functionality has yet to be implemented.

peer/peer.go
bfbachmann: `AddPeer` handles the case where the peer is already in our subnet. I think your description of what `Receive` should do hits the nail on the head.

peer/peer.go
bfbachmann: Yea I guess this should be called `sendPeerInfo`. It should only get called when a peer requests peer info.

subnet/subnet.go
bfbachmann: Wow yea that is just disgusting.

peer/peer.go
bfbachmann: If it's capitalized its public, so I decided against it.

PR #41, opened 2017-05-27
31 Creating and Handling P2P Communications

bfbachmann:
# What Changed:
- Updated peers to handle asynchronous requests and responses
- Added some convenience methods to `message` 
- Added `Peer` and `Subnet` tests


errors/error.go
```diff
@@ -0,0 +1,58 @@
+package errors
+
+import "fmt"
+
+const (
```
jordanschalm: Modify the error system so that error codes have general meaning and messages are inserted in the constructor and contain info specific to the conditions of the error.
Like `errors.New(Code, "asdfasdf")

errors/error.go
```diff
@@ -0,0 +1,58 @@
...
+		msg = subnetFullMsg
+		break
+	case InvalidResourceType:
+		msg = invalidResourceTypeMsg
+		break
+	case NotImplemented:
+		msg = notImplementedMsg
+		break
+	default:
+		panic("Attempt to create error with invalid error code")
```
jordanschalm: 😮

peer/peer.go
```diff
@@ -1,36 +1,45 @@
...
+	"github.com/libp2p/go-libp2p-net"
 	lpeer "github.com/libp2p/go-libp2p-peer"
 	pstore "github.com/libp2p/go-libp2p-peerstore"
-	swarm "github.com/libp2p/go-libp2p-swarm"
+	"github.com/libp2p/go-libp2p-swarm"
 	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
 	ma "github.com/multiformats/go-multiaddr"
+	protoerr "github.com/ubclaunchpad/cumulus/errors"
+	"github.com/ubclaunchpad/cumulus/message"
+	sn "github.com/ubclaunchpad/cumulus/subnet"
```
jordanschalm: Why override `subnet`, `peerstore`, `multiaddr`

peer/peer.go
```diff
@@ -1,36 +1,45 @@
...
 	DefaultIP = "127.0.0.1"
+	// Timeout is the time after which reads from a stream will timeout
+	Timeout = time.Second * 30
 )
 
 // Peer is a cumulus Peer composed of a host
 type Peer struct {
 	host.Host
+	subnet        sn.Subnet
+	listeners     map[string]chan message.Response
```
jordanschalm: Shouldn't `listeners` and its lock be per-stream? Our peer will be connected to multiple other peers and we should partition the listeners based on which stream they're listening on

peer/peer.go
```diff
@@ -69,9 +78,16 @@ func New(ip string, port int) (*Peer, error) {
 		return nil, err
 	}
 
+	subnet := *sn.New(sn.DefaultMaxPeers)
+
 	// Actually create the host and peer with the network we just set up.
 	host := bhost.New(netwrk)
```
jordanschalm: What do you have against `o`?

peer/peer.go
```diff
@@ -84,43 +100,204 @@ func New(ip string, port int) (*Peer, error) {
...
 func (p *Peer) Receive(s net.Stream) {
-	log.Debug("Setting basic stream handler.")
-	p.doCumulus(s)
-}
+	// Get remote peer's full multiaddress
+	remoteMA, err := NewMultiaddr(
+		s.Conn().RemoteMultiaddr(), s.Conn().RemotePeer())
+	if err != nil {
+		log.WithError(err).Error(
+			"Failed to obtain valid remote peer multiaddress")
```
jordanschalm: Put this on one line

peer/peer.go
```diff
@@ -84,43 +100,204 @@ func New(ip string, port int) (*Peer, error) {
...
 
-// Communicate with peers.
-// TODO: Update this to do something useful. For now it just reads from the
-// stream and writes back what it read.
-func (p *Peer) doCumulus(s net.Stream) {
-	buf := bufio.NewReader(s)
-	str, err := buf.ReadString('\n')
+	// Add the remote peer to this peer's subnet
+	err = p.subnet.AddPeer(remoteMA, s)
 	if err != nil {
```
jordanschalm: If the error always means `full` then change it to a boolean

peer/peer.go
```diff
@@ -84,43 +100,204 @@ func New(ip string, port int) (*Peer, error) {
...
 		return
 	}
+	go p.Listen(remoteMA, s)
+}
+
+// Connect adds the given multiaddress to p's Peerstore and opens a stream
+// with the peer at that multiaddress if the multiaddress is valid, otherwise
+// returns error. On success the stream and corresponding multiaddress are
+// added to this peer's subnet.
+func (p *Peer) Connect(peerma string) (net.Stream, error) {
```
jordanschalm: Can we just call multiaddresses `addr`'s always? `ma` and all this is confusing

peer/peer.go
```diff
@@ -84,43 +100,204 @@ func New(ip string, port int) (*Peer, error) {
...
+
+	// Store the peer's address in this host's PeerStore
+	p.Peerstore().AddAddr(peerid, targetAddr, pstore.PermanentAddrTTL)
+
+	// Open a stream with the peer
+	stream, err := p.NewStream(context.Background(), peerid, CumulusProtocol)
+	if err != nil {
+		return nil, err
+	}
+	err = stream.SetDeadline(time.Now().Add(Timeout))
```
jordanschalm: If we're setting timeouts then we need to have ping/pong messages right?

peer/peer.go
```diff
@@ -84,43 +100,204 @@ func New(ip string, port int) (*Peer, error) {
...
+		return nil, err
+	}
+
+	err = p.subnet.AddPeer(mAddr, stream)
+	if err != nil {
+		stream.Close()
+	}
+
+	go p.Listen(mAddr, stream)
+	return stream, nil
```
jordanschalm: Where are the streams stored? We need a way to send to a particular stream or send to all streams given the Peer struct.

peer/peer.go
```diff
@@ -84,43 +100,204 @@ func New(ip string, port int) (*Peer, error) {
...
+func (p *Peer) Broadcast(m message.Message) error {
+	return errors.New("Function not implemented")
+}
+
+// Request sends a request to a remote peer over the given stream.
+// Returns response if a response was received, otherwise returns error.
+// You should typically run this function in a goroutine
+func (p *Peer) Request(req message.Request, s net.Stream) (*message.Response, error) {
+	// Set up a listen channel to listen for the response (remove when done)
+	lchan := p.newListener(req.ID)
```
jordanschalm: I don't like this `newListener` business. I think we should partition better based on stream

peer/peer.go
```diff
@@ -84,43 +100,204 @@ func New(ip string, port int) (*Peer, error) {
...
+		return &res, nil
+	case <-time.After(Timeout):
+		return nil, errors.New("Timed out waiting for response")
+	}
+}
 
-	log.Debugf("Peer %s read: %s", p.ID(), str)
+// Respond responds to a request from another peer
+func (p *Peer) Respond(req message.Request, s net.Stream) {
+	response := message.Response{ID: req.ID, Error: nil, Resource: nil}
```
jordanschalm: Don't need to explicitly set Err/REs to nil

peer/peer.go
```diff
@@ -84,43 +100,204 @@ func New(ip string, port int) (*Peer, error) {
...
+}
+
+// Listen listens for messages over the stream and responds to them, closing
+// the given stream and removing the remote peer from this peer's subnet when
+// done. This should be run as a goroutine.
+func (p *Peer) Listen(remoteMA ma.Multiaddr, s net.Stream) {
+	defer s.Close()
+	defer p.subnet.RemovePeer(remoteMA)
+	for {
+		err := s.SetDeadline(time.Now().Add(Timeout))
```
jordanschalm: This will require heartbeats

subnet/subnet.go
```diff
@@ -0,0 +1,83 @@
+package subnet
+
+import (
+	"errors"
+
+	log "github.com/Sirupsen/logrus"
+	net "github.com/libp2p/go-libp2p-net"
```
jordanschalm: Don't need this it will be called net

subnet/subnet.go
```diff
@@ -0,0 +1,83 @@
...
+
+const (
+	// DefaultMaxPeers is the maximum number of peers a peer can be in
+	// communication with at any given time.
+	DefaultMaxPeers = 6
+)
+
+// Subnet represents the set of peers a peer is connected to at any given time
+type Subnet struct {
+	peers    map[ma.Multiaddr]net.Stream
```
jordanschalm: 👍

jordanschalm (changes requested):
Good but I think the multiplexing of messages to goroutines needs some work.

peer/peer.go
bfbachmann: Yea haven't got around to implementing that yet.

PR #46, opened 2017-06-03
45 cobra cli

jordanschalm:


cmd/send.go
```diff
@@ -0,0 +1,52 @@
+// Copyright © 2017 NAME HERE <EMAIL ADDRESS>
```
bfbachmann: I don't know if we actually need this.

chadlagore (approved):
🥇 ⭐️

bfbachmann (approved):

PR #52, opened 2017-06-03
Validation Refactor and Error Codes

chadlagore:
**Status** Ready for review.

# What Changed

🗺 Refactored the validation functions into their own file (along with tests).
⚙️ Added error codes to enforce proper failing in test suite.

## TODO

- [x] Finish refactor (make it easy to swap out `[]*Transactions` data structure.
- [x] ~Test `bc.GetInputTransaction` method.~

### DIFF
--- a/blockchain/block.go
+++ b/blockchain/block.go
@@ -73,11 +73,11 @@ func DecodeBlock(r io.Reader) *Block {
 
 // ContainsTransaction returns true and the transaction itself if the Block
 // contains the transaction.
-func (b *Block) ContainsTransaction(t *Transaction) (bool, *Transaction) {
-	for _, tr := range b.Transactions {
+func (b *Block) ContainsTransaction(t *Transaction) (bool, uint32) {
+	for i, tr := range b.Transactions {
 		if HashSum(t) == HashSum(tr) {
-			return true, tr
+			return true, uint32(i)
 		}
 	}
-	return false, nil
+	return false, 0
 }
--- a/blockchain/transaction.go
+++ b/blockchain/transaction.go
@@ -16,6 +16,7 @@ const (
 type TxHashPointer struct {
 	BlockNumber uint32
 	Hash        Hash
+	Index       uint32
 }
 
 // Marshal converts a TxHashPointer to a byte slice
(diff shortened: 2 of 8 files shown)


blockchain/block.go
```diff
@@ -73,11 +73,11 @@ func DecodeBlock(r io.Reader) *Block {
 
 // ContainsTransaction returns true and the transaction itself if the Block
 // contains the transaction.
-func (b *Block) ContainsTransaction(t *Transaction) (bool, *Transaction) {
-	for _, tr := range b.Transactions {
+func (b *Block) ContainsTransaction(t *Transaction) (bool, uint32) {
```
jordanschalm, line 76: Why `uint32`?

blockchain/transaction.go
```diff
@@ -88,3 +89,23 @@ func (t *Transaction) Marshal() []byte {
 	buf = append(buf, t.Sig.Marshal()...)
 	return buf
 }
+
+// InputsEqualOutputs returns true if t.Inputs == other.Outputs, as well
+// as the difference between the two (outputs - inputs).
+func (t *Transaction) InputsEqualOutputs(other ...*Transaction) (bool, int) {
```
jordanschalm: Checking that the addresses match is done somewhere else?

jordanschalm: Also, why are we returning the diff as well?

blockchain/validation.go
```diff
@@ -0,0 +1,105 @@
...
+
+// TransactionCode is returned from ValidTransaction.
+type TransactionCode uint32
+
+// BlockCode is returned from ValidBlock.
+type BlockCode uint32
+
+const (
+	// ValidTransaction is returned when transaction is valid.
+	ValidTransaction TransactionCode = 0
```
jordanschalm: Use `iota` (https://golang.org/ref/spec#Iota)

blockchain/validation.go
```diff
@@ -0,0 +1,105 @@
...
+	// Overspend is returned when transaction outputs exceed transaction inputs.
+	Overspend TransactionCode = 2
+	// BadSig is returned when the signature verification fails.
+	BadSig TransactionCode = 3
+	// Respend is returned when inputs have been spent elsewhere in the chain.
+	Respend TransactionCode = 4
+)
+
+const (
+	// ValidBlock is returned when the block is valid.
```
jordanschalm: Same -- use iota

blockchain/validation_test.go
```diff
@@ -0,0 +1,157 @@
+package blockchain
+
+import (
+	crand "crypto/rand"
```
jordanschalm: Just use `rand` if we aren't distinguishing between math/rand and crypto/rand

jordanschalm (approved):
👍

blockchain/block.go
chadlagore: Good question. `uint32` for blocks because thats baked into the data structure, avoid overflow if we have billions of blocks. Thought we might as well keep the same datatype for transactions as well, no such thing as a negative transaction index.

blockchain/transaction.go
chadlagore: In case we want to verify that the diff is equal to some fee.

jordanschalm: If we aren't using it, I'd leave it out until we need it. Fees will eventually be part of the transaction so we can verify as part of this function anyway right?

PR #61, opened 2017-06-07
Model for peer connections

jordanschalm:


peer/peer.go
```diff
@@ -1,174 +1,268 @@
...
-	// Now we can build a full multiaddress to reach this host
-	// by encapsulating both addresses:
-	fullAddr := addr.Encapsulate(hostAddr)
-	log.Info("I am ", fullAddr)
+// Addrs returns the list of addresses of the peers in the peerstore in the form
+// <IP addr>:<port>
+func (ps *PeerStore) Addrs() []string {
+	ps.lock.RLock()
+	defer ps.lock.RUnlock()
+	addrs := make([]string, len(ps.peers), len(ps.peers))
```
jordanschalm, line 69: Don't need the third argument `cap` unless it's greater than second argument.

peer/peer.go
```diff
@@ -1,174 +1,268 @@
...
+		case push = <-p.pushChan:
+			break
+		case <-time.After(Timeout):
+			continue
+		}
+
+		switch push.ResourceType {
+		case message.ResourcePeerInfo:
+			for _, addr := range push.Resource.([]string) {
+				conn.Listen(addr, func(c net.Conn) {
```
jordanschalm: What the heck?

bfbachmann: Yup this is one of those bugs I was telling you about..... just garbage.

peer/peer.go
bfbachmann: Irrelevant now that I switched the peerstore back to a map. Didn't know we could iterate over a map back then.... my bad.

PR #64, opened 2017-06-10
Transaction Pooling

chadlagore:
**Status**: Ready to merge. 

# What Changed

+ Wrote tests and function specifications for transaction pooling.
+ Exported `NewWallet` from the `blockchain` package for easy `Wallet` creation.
+ Added `ordered_map` package.

## How To Test Drive This PR

`go test` and watch em fail! :tada:

## TODO

- [x] Implement core pooling functionality.


blockchain/pool.go
```diff
@@ -0,0 +1,48 @@
+package blockchain
```
jordanschalm: I think we should probably move this into its own package

blockchain/pool.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// GetTransaction returns the transctions with input transaction Hash h.
+func (p *Pool) GetTransactions(h Hash) []*Transaction {
+	return nil
+}
+
+// PutTransaction inserts a transaction into the pool, returning
+// true if the Transaction was inserted (was valid).
+func (p *Pool) PutTransaction(t *Transaction) bool {
```
jordanschalm: Replace with the channel

blockchain/pool.go
```diff
@@ -0,0 +1,48 @@
...
+	return nil
+}
+
+// Len returns the number of transactions in the Pool.
+func (p *Pool) Len() int {
+	return 0
+}
+
+// GetTransaction returns the transctions with input transaction Hash h.
+func (p *Pool) GetTransactions(h Hash) []*Transaction {
```
jordanschalm: should only be one transaction per input

blockchain/pool.go
```diff
@@ -0,0 +1,48 @@
...
+
+// PutTransaction inserts a transaction into the pool, returning
+// true if the Transaction was inserted (was valid).
+func (p *Pool) PutTransaction(t *Transaction) bool {
+	return false
+}
+
+// RemoveTransaction removes a transaction from the Pool, returning
+// true if the Transaction existed in the pool.
+func (p *Pool) RemoveTransaction(t *Transaction) bool {
```
jordanschalm: don't know if we need this

jordanschalm (commented):
Lookin' good

blockchain/pool.go
chadlagore: For updating when we see new blocks.

blockchain/pool.go
chadlagore: Hmm, I still need to put a transaction into the `Pool` once it pops out of the channel, no?

pool/pool.go
```diff
@@ -0,0 +1,87 @@
...
+		p.OrderReverseMap[p.Len()] = t.Input.Hash
+		p.Transactions[t.Input.Hash] = t
+	}
+	return ok
+}
+
+// Delete removes a transaction from the Pool.
+func (p *Pool) Delete(t *blockchain.Transaction) {
+	i := p.OrderMap[t.Input.Hash]
+	if i < p.Len() {
```
jordanschalm: Doesn't this also need to shuffle transaction hash indices
Ex: you have `{1: a, 2: b, 3: c, 4: d, 5: e}` and remove 3 to get `{1: a, 2: b, 4: d, 5: e}`. But now d isn't the 4th element anymore it's the 3rd, same for e. Does this matter?

pool/pool.go
```diff
@@ -0,0 +1,87 @@
...
+		p.OrderMap[t.Input.Hash] = p.Len()
+		p.OrderReverseMap[p.Len()] = t.Input.Hash
+		p.Transactions[t.Input.Hash] = t
+	}
+	return ok
+}
+
+// Delete removes a transaction from the Pool.
+func (p *Pool) Delete(t *blockchain.Transaction) {
+	i := p.OrderMap[t.Input.Hash]
```
jordanschalm: Can also do:
```
i, ok := p.orderMap[t.input.Hash]
if ok { ... }
```

`ok` is true when the map contains the key

pool/pool.go
```diff
@@ -0,0 +1,87 @@
...
+	for _, t := range b.Transactions {
+		p.Delete(t)
+	}
+	return true
+}
+
+// GetTxns returns the the largest of l or size of pool transactions.
+// It selects the highest priority transactions.
+func (p *Pool) GetTxns(l int) []*blockchain.Transaction {
+	var txns []*blockchain.Transaction
```
jordanschalm: How about:
```
if p.Len() < l {
  l = p.Len()
}
txns := make([]*blockchain.Transaction, l)
for i := 0; i < l; i++ {
  txns[i] = p.GetN(i)
}
```

jordanschalm: We probably want an empty slice, rather than nil for an empty list of transactions

jordanschalm (commented):
👍  Awesome

pool/pool.go
chadlagore: In that case, `OrderReverseMap` still reflects `{a: 1, b: 2, d: 4, e: 5}`.

pool/pool.go
```diff
@@ -0,0 +1,125 @@
...
+		ValidTransactions: map[blockchain.Hash]*PooledTransaction{},
+	}
+}
+
+// Len returns the number of transactions in the Pool.
+func (p *Pool) Len() int {
+	return len(p.ValidTransactions)
+}
+
+// Get returns the transction with input transaction Hash h.
```
jordanschalm: a wee typo

pool/pool.go
```diff
@@ -0,0 +1,125 @@
...
+	return p.ValidTransactions[h].Transaction
+}
+
+// GetN returns the Nth transaction in the pool.
+func (p *Pool) GetN(N int) *blockchain.Transaction {
+	return p.Order[N].Transaction
+}
+
+// GetIndex returns the index of the transaction in the ordering.
+func (p *Pool) GetIndex(t *blockchain.Transaction) int {
```
jordanschalm, line 45: 👍

pool/pool.go
```diff
@@ -0,0 +1,126 @@
...
+	p.Order = append(p.Order, vt)
+	p.ValidTransactions[t.Input.Hash] = vt
+}
+
+// Delete removes a transaction from the Pool.
+func (p *Pool) Delete(t *blockchain.Transaction) {
+	vt, ok := p.ValidTransactions[t.Input.Hash]
+	if ok {
+		i := p.GetIndex(vt.Transaction)
+		p.Order = append(p.Order[0:i], p.Order[i:p.Len()-1]...)
```
jordanschalm: This should be `p.Order[i+1:p.Len()]...)` shouldn't it? Don't include `i` on the low end and don't include `p.Len` on the high end? Same as python

jordanschalm (approved):
👍

PR #66, opened 2017-06-10
25 mining proof of work

david-julien:
Initial mining proof of work code, not all tests written yet


blockchain/hash.go
```diff
@@ -1,9 +1,38 @@
 package blockchain
```
jordanschalm, line 1: Move all the miner stuff into its own package

blockchain/hash.go
```diff
@@ -1,9 +1,38 @@
...
+	"log"
+)
 
-// HashLen is the length in bytes of a hash.
-const HashLen = 32
+const (
+	// HashLen is the length in bytes of a hash.
+	HashLen = 32
+	// MaxDifficultyHex is the maximum difficulty value represented as a hex string
+	MaxDifficultyHex = "000000FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF"
```
jordanschalm: Can this be replace by using `math.big`?

blockchain/hash.go
```diff
@@ -26,3 +55,43 @@ type Marshaller interface {
 func HashSum(m Marshaller) Hash {
 	return sha256.Sum256(m.Marshal())
 }
+
+// CompareTo compares two hashes, it returns true if the operation of the first hash on the second hash specified by the comparator is true, and false otherwise
+func CompareTo(h1 Hash, h2 Hash, comparator int) bool {
```
jordanschalm: We will only ever be checking for hashes being less than right? Can we simplify this accordingly and maybe remove those comparator constants unless we need arbitrary comparisons?

blockchain/hash.go
```diff
@@ -26,3 +55,43 @@ type Marshaller interface {
...
+			return 1 == comparator
+		} else if h1[i] < h2[i] {
+			return -1 == comparator
+		}
+	}
+	return 0 == comparator
+}
+
+// HexStringToHash converts a big endian hex string to a hash
+func HexStringToHash(s string) Hash {
```
jordanschalm: Can this be replaced by `math.big`?

blockchain/miner.go
```diff
@@ -0,0 +1,103 @@
...
+// DoubleHashSum computes the hash 256 of the marshalled mining header twice
+func (mh *MiningHeader) DoubleHashSum() Hash {
+	hash := sha256.Sum256(mh.Marshal())
+	hash = sha256.Sum256(hash[:])
+	return hash
+}
+
+// Mine continuously increases the nonce and tries to verify the proof of work until the puzzle is solved
+func (mh *MiningHeader) Mine() bool {
+	if mh.VerifyMiningHeader() == false {
```
jordanschalm: use `!` rather than comparing to false

blockchain/miner.go
```diff
@@ -0,0 +1,103 @@
...
+		return false
+	}
+
+	for mh.VerifyProofOfWork() == false {
```
jordanschalm: same here

blockchain/miner.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if CompareTo(mh.Target, MinHash, EqualTo) || CompareTo(mh.Target, MaxDifficulty, GreaterThan) {
+		log.Error("Invalid target in mining header")
+		return false
+	}
+	return true
+}
+
+// AppendUint32ToSlice converts a uint32 to a byte slice and appends it to a byte slice
+func AppendUint32ToSlice(s *[]byte, num uint32) {
```
jordanschalm: Do we need this function? (and below) seems like they're only called once

blockchain/miner.go
```diff
@@ -0,0 +1,103 @@
...
+	buf = append(buf, mh.LastBlock.Marshal()...)
+	buf = append(buf, mh.RootHash.Marshal()...)
+	buf = append(buf, mh.Target.Marshal()...)
+	AppendUint32ToSlice(&buf, mh.Time)
+	AppendUint64ToSlice(&buf, mh.Nonce)
+	return buf
+}
+
+// SetMiningHeader sets the mining header
+func (mh *MiningHeader) SetMiningHeader(lastBlock Hash, rootHash Hash, target Hash) {
```
jordanschalm: Rather than a setter can we just use composite instantiation? What does the setter buy us?

ie. `mh := MiningHeader{lastBlock, rootHash, target}`

jordanschalm (approved):
Really good! I think there are some more simplifications we can add

bfbachmann (commented):
Great job 🥇

blockchain/hash.go
david-julien: yes we need all of them for miningheader verification purposes before mining, we'll see if i can clean this up when we start using math.big

blockchain/miner.go
david-julien: it sets the time to the time variable to the current time in the miningheader struct, not quite sure how to set default values in golang structs

blockchain/hash.go
david-julien: math.big uses the exact same arbitrary constants for comparison, so if we wanted to compare two hashes we would have to convert to math.big and then use the same constants.

https://golang.org/src/math/big/float.go?s=43528:43561#L1637

blockchain/miner.go
jordanschalm: The default value is the zero value for the type. You can do something like:
```
mh := MiningHeader{
  LastBlock: lastBlock,
  ...
  timeField: time.Now(),
}

File: blockchain/hash.go [hunk 9]
```diff
@@ -26,3 +56,63 @@ type Marshaller interface {
...
+		h[i] = bytes[HashLen-1-i]
+	}
+
+	return h
+}
+
+// HexString returns the hexadecimal representation of the hash
+func (h Hash) HexString() string {
+	var hashBigEndian Hash
+	for i := 0; i < HashLen; i++ {
```
chadlagore: I notice this pattern of inverting the hex string a couple of times, perhaps a helper function?

blockchain/hash_test.go
```diff
@@ -0,0 +1,160 @@
+package blockchain
+
+import (
+	"testing"
+)
+
+func TestParseString(t *testing.T) {
```
chadlagore: :clap: Great coverage. But I wonder if `ParseString` couldn't be broken out into a set of helper functions,  with a test for each so that when something breaks we have a clearer idea of what part broke?

miner/miner.go
```diff
@@ -0,0 +1,94 @@
...
+	mh.LastBlock = lastBlock
+	mh.RootHash = rootHash
+	mh.Target = target
+	mh.Time = uint32(time.Now().Unix())
+	mh.Nonce = 0
+}
+
+// VerifyProofOfWork computes the hash of the MiningHeader and returns true if the result is less than the target
+func (mh *MiningHeader) VerifyProofOfWork() bool {
+	if mh.DoubleHashSum().CompareTo(mh.Target, bc.LessThan) {
```
chadlagore: Non-blocking, but in terms of convention we should consider the use of `bc` as a namespace vs a concrete instance of a `Blockchain`. I've been using it as the later, but I'm game for whatever as long as we're consistent.

blockchain/hash_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if h.ParseString("1") != nil {
+		t.Fail()
+	}
+
+	if h.ParseString("a") != nil {
+		t.Fail()
+	}
+
+	if h.ParseString("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff") != nil {
```
jordanschalm: @david-julien Is it possible to replace these huge hex strings with `math.big`. I don't like putting big string literal hexes all over the code.

Rather than doing `hash.ParseString()` to get a hash from a big number, can we use the `math.big` exponentiation function like `z.Exp(2, 50, 0)` to get 2^50.

If we need the ParseString function, do `func ParseString(s string) Hash` rather than having the hash as the receiver.

blockchain/hash.go
```diff
@@ -26,3 +56,63 @@ type Marshaller interface {
...
+			return GreaterThan == comparator
+		} else if h[i] < h2[i] {
+			return LessThan == comparator
+		}
+	}
+	return EqualTo == comparator
+}
+
+// ParseString sets the hash to the the value represented by a hexadecimal string, and returns the hash
+func (h *Hash) ParseString(s string) *Hash {
```
jordanschalm: I'd like to remove the need for hex strings if possible. See comment below

miner/miner.go
chadlagore: Also here the namespacing does not really tell me what `LessThan` is. I wonder if 

+ `hash.GreaterThan(otherHash)`
+ `hash.LessThan(otherHash)`
+ `hash.EqualTo(otherHash)`

is a more clear API?

jordanschalm (approved):
Awesome, if we can remove dependence on literal hex strings and use math.big instead that would be awesome, otherwise ready to merge with some more test coverage.

chadlagore (approved):
This is great 🥇 Nothing blocking but I think `ParseString` could be broken out into helper functions, and the API around `CompareTo` could be a little more clear. Up to you for changes, otherwise :+1:.

blockchain/hash_test.go
chadlagore: Was trying to articulate this very thing.

miner/miner.go
bfbachmann: You could make this a one liner by hitting a little:
```
return mh.DoubleHashSum().CompareTo(mh.Target, bc.LessThan)
```

miner/miner.go
```diff
@@ -0,0 +1,94 @@
+package miner
```
chadlagore, line 1: Can we introduce an interface for `MiningHeader` or maybe `Miner`? I can imagine a scenario where we want to do an integration test where we mine a block (without waiting).

bfbachmann (approved):
🥇

PR #80, opened 2017-06-25
Implementing Push and Request Logic

chadlagore:
**Status** _open for visibility._

# What Changed

+ Added `Worker`, `Work`, and a dispatcher.
+ Currently writing tests.

## How to Test Drive

+ Currently `go test`, more to come...

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -53,23 +53,8 @@ func Run(cfg conf.Config) {
 		}
 	}()
 
-	// If a target peer was supplied, connect to it and try discover and connect
-	// to its peers
-	if len(cfg.Target) > 0 {
-		peerInfoRequest := msg.Request{
-			ID:           uuid.New().String(),
-			ResourceType: msg.ResourcePeerInfo,
-		}
-
-		log.Infof("Dialing target %s", cfg.Target)
-		c, err := conn.Dial(cfg.Target)
-		if err != nil {
-			log.WithError(err).Fatalf("Failed to connect to target")
-		}
-		peer.ConnectionHandler(c)
-		p := peer.PStore.Get(c.RemoteAddr().String())
-		p.Request(peerInfoRequest, peer.PeerInfoHandler)
-	}
+	// Connect to the target and discover its peers.
+	ConnectAndDiscover(cfg.Target)
 
 	// Try maintain as close to peer.MaxPeers connections as possible while this
 	// peer is running
@@ -83,6 +68,25 @@ func Run(cfg conf.Config) {
 	select {}
 }
 
+// ConnectAndDiscover tries to connect to a target and discover its peers.
+func ConnectAndDiscover(target string) {
+	if len(target) > 0 {
+		peerInfoRequest := msg.Request{
+			ID:           uuid.New().String(),
+			ResourceType: msg.ResourcePeerInfo,
+		}
+
+		log.Infof("Dialing target %s", target)
+		c, err := conn.Dial(target)
+		if err != nil {
+			log.WithError(err).Fatalf("Failed to connect to target")
+		}
+		peer.ConnectionHandler(c)
+		p := peer.PStore.Get(c.RemoteAddr().String())
+		p.Request(peerInfoRequest, peer.PeerInfoHandler)
+	}
+}
+
 // RequestHandler is called every time a peer sends us a request message except
 // on peers whos PushHandlers have been overridden.
 func RequestHandler(req *msg.Request) msg.Response {
@@ -92,16 +96,10 @@ func RequestHandler(req *msg.Request) msg.Response {
 	case msg.ResourcePeerInfo:
 		res.Resource = peer.PStore.Addrs()
 	case msg.ResourceBlock:
-		// Unmarshal request.
 		work := BlockWork{}
-		// Define callback
-		// Add to BlockWorkQueue
 		BlockWorkQueue <- work
 	case msg.ResourceTransaction:
-		// Unmarshal request.
 		work := TransactionWork{}
-		// Define callback.
-		// Add to TransactionWorkQueue
 		TransactionWorkQueue <- work
 	default:
 		res.Error = msg.NewProtocolError(msg.InvalidResourceType,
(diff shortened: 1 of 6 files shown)


app/app.go
```diff
@@ -36,9 +44,12 @@ func Run(cfg conf.Config) {
 	log.Infof("Starting listener on %s:%d", cfg.Interface, cfg.Port)
 	peer.LocalAddr = fmt.Sprintf("%s:%d", cfg.Interface, cfg.Port)
 	go func() {
-		err := conn.Listen(fmt.Sprintf("%s:%d", cfg.Interface, cfg.Port), peer.ConnectionHandler)
+		address := fmt.Sprintf("%s:%d", cfg.Interface, cfg.Port)
```
bfbachmann, line 47: Might as well move this line up above this goroutine and use `address` in the calls to `log.Infof` on line 44 and `Fatalf` on like 52.

app/app.go
```diff
@@ -63,20 +74,34 @@ func Run(cfg conf.Config) {
...
+	// Request the blockchain.
+	log.Info("Requesting blockchain from peers... ")
+	RequestBlockChain()
+
+	// Return to command line.
 	select {}
 }
 
-// RequestHandler is called every time a peer sends us a request message expect
+// RequestHandler is called every time a peer sends us a request message except
```
bfbachmann: sad

app/work.go
```diff
@@ -0,0 +1,37 @@
+package app
+
+import "github.com/ubclaunchpad/cumulus/blockchain"
+
+const (
+	// BlockQueueBuffer is the size of the BlockQueue channel.
+	BlockQueueBuffer = 100
```
bfbachmann: I would call this BlockQueueSize. Same for TransactionQueueBuffer.

bfbachmann (commented):
Looks good, I just don't get why we need to have distributed workers and a job dispatcher. Won't these jobs just take a few seconds at most? I figured we could implement the same functionality with just a simple PushHandler and RequestHandler that do what HandleTransaction and HandleBock do without needing to worry about queueing jobs and creating workers etc.

chadlagore:
@bfbachmann We should benchmark it, but `ValidTransaction` and `ValidBlock` are actually quite expensive, especially if we choose to not cap the block size. I see your point, but I'm not convinced that this is premature optimization yet.

app/app.go
```diff
@@ -85,17 +111,56 @@ func RequestHandler(req *msg.Request) msg.Response {
...
 	case msg.ResourceTransaction:
+		work := TransactionWork{}
+		TransactionWorkQueue <- work
 	default:
 		// Invalid resource type. Ignore
 	}
+}
+
+// Shutdown shuts down the node (save the pool and the chain to disk).
+func Shutdown() {}
```
jordanschalm: Saving to disk might not be a prototype feature

app/app.go
```diff
@@ -85,17 +111,56 @@ func RequestHandler(req *msg.Request) msg.Response {
...
+// incoming messages.
+func initializeNode() {
+	tpool = pool.New()
+	intializeQueues()
+	initializeWorkers()
+}
+
+// intializeQueues makes all necessary queues.
+func intializeQueues() {
+	BlockWorkQueue = make(chan BlockWork, BlockQueueBuffer)
```
jordanschalm: Maybe `BlockBufferSize` or something to make it more clear that it's a size

app/work.go
```diff
@@ -0,0 +1,37 @@
...
+	// TransactionQueueBuffer is the size of the BlockQueue channel.
+	TransactionQueueBuffer = 100
+)
+
+// Responder is used to handle requests who require a response.
+type Responder interface {
+	Send(ok bool)
+	Lock()
```
jordanschalm, line 15: Why lock/unlock? Why does send only take a bool?

app/worker.go
```diff
@@ -0,0 +1,91 @@
...
+func (w GenericWorker) Start() {
+	go func() {
+		for {
+			// Wait for work.
+			log.WithFields(log.Fields{
+				"id": w.ID,
+			}).Debug("Worker waiting for work.")
+			select {
+			case work := <-TransactionWorkQueue:
+				log.WithFields(log.Fields{
```
jordanschalm: You can avoid this by adding a `log` field to the worker and using 
```
worker := Worker{
  log: log.WithField("id", id),
  ...
}
```

jordanschalm: The field will persist on every subsequent `log.Whatever` call

app/worker.go
```diff
@@ -0,0 +1,91 @@
...
+				return
+			}
+		}
+	}()
+}
+
+// NB: We're currently imposing a validation layer at the app level
+// using methods like SetUnsafe to get the transaction into the
+// pool, etc. Figure out a nicer way to do this, ie: where *should*
+// we validate?
```
jordanschalm: I thought `pool` would have a channel of incoming, unvalidated transactions that `app` could push onto. `pool` should do all *transaction* validation

app/worker.go
```diff
@@ -0,0 +1,91 @@
...
+// HandleTransaction handles new instance of TransactionWork.
+func (w *GenericWorker) HandleTransaction(work TransactionWork) {
+	ok, _ := chain.ValidTransaction(work.Transaction)
+	if ok {
+		tpool.SetUnsafe(work.Transaction)
+	}
+
+	// Respond to the request if a response method was provided.
+	if work.Responder != nil {
+		work.Responder.Lock()
```
jordanschalm, line 66: Minor thing but idiom is:
```
x.Lock()
defer x.Unlock()
...
```

Rationale is if the body ever becomes more complex your `defer` will handle it without needing to think about it

jordanschalm (approved):
Looks good. Couple of minor comments. It does seem quite complex, but maybe that will be necessary.

app/work.go
chadlagore: `Send` is sort of prototype-y. Not sure what message may need to go back to the app.

chadlagore: Threads may share a `Responder`.

app/worker.go
chadlagore: WHAT

app/worker.go
chadlagore: I think so too. I'm would be doing transaction validation at the `pool` level and block validation at the `app` level, it ust felt weird. I'll change it back.

app/worker.go
chadlagore: Yeah, I'm game thx.

app/app.go
chadlagore: Totally 🤦‍♂️

app/work.go
chadlagore: Totally 🤦‍♀️

app/app.go
chadlagore: ✅

app/app.go
chadlagore: I think `Run` will get a big overhaul yet.

app/worker.go
jordanschalm: Just need to do `worker.log.Whatever` rather than `log.Whatever`

app/worker.go
chadlagore: I wonder if it makes sense to have the blockchain make those choices too. ie `AppendBlock` simply fails if the block is invalid.

chadlagore: It makes no sense to append a non-legit block.

PR #81, opened 2017-07-01
75 creating spending mined coins

david-julien:
Still need to finish writing tests for consensus.

### DIFF
--- a/blockchain/validation.go
+++ b/blockchain/validation.go
@@ -1,7 +1,11 @@
 package blockchain
 
 // ValidTransaction checks whether a transaction is valid, assuming the
-import "crypto/ecdsa"
+import (
+	"crypto/ecdsa"
+	"math/big"
+	"time"
+)
 
 // TransactionCode is returned from ValidTransaction.
 type TransactionCode uint32
@@ -28,6 +32,10 @@ const (
 	// BadTransaction is returned when the block contains an invalid
 	// transaction.
 	BadTransaction BlockCode = iota
+	// BadTime is returned when the block contains an invalid time
+	BadTime BlockCode = iota
+	// BadTarget is returned when the block contains an invalid target
+	BadTarget BlockCode = iota
 	// BadBlockNumber is returned when block number is not one greater than
 	// previous block.
 	BadBlockNumber BlockCode = iota
@@ -36,6 +44,13 @@ const (
 	// DoubleSpend is returned when two transactions in the block share inputs,
 	// but outputs > inputs.
 	DoubleSpend BlockCode = iota
+)
+
+var (
+	// MinDifficulty is the minimum difficulty
+	MinDifficulty = new(big.Int).Sub(BigExp(2, 232), big.NewInt(1))
+	// MaxDifficulty is the maximum difficulty value
+	MaxDifficulty = big.NewInt(1)
 )
 
 // ValidTransaction tests whether a transaction valid.
@@ -90,6 +105,17 @@ func (bc *BlockChain) ValidBlock(b *Block) (bool, BlockCode) {
 		}
 	}
 
+	// Check that the target is within the min and max difficulty levels
+	target := HashToBigInt(b.Target)
+	if target.Cmp(MinDifficulty) == 1 || target.Cmp(MaxDifficulty) == -1 {
+		return false, BadTarget
+	}
+
+	// Check that time is not greater than current time or equal to 0
+	if uint32(b.Time) == 0 || uint32(b.Time) > uint32(time.Now().Unix()) {
+		return false, BadTime
+	}
+
 	// Check that hash of last block is correct
 	if HashSum(lastBlock) != b.LastBlock {
 		return false, BadHash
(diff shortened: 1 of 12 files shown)


blockchain/validation.go
```diff
@@ -69,8 +121,51 @@ func (bc *BlockChain) ValidTransaction(t *Transaction) (bool, TransactionCode) {
 	return true, ValidTransaction
 }
 
+// ValidCloudBase returns true if a transaction is a valid CloudBase transaction
+// and false otherwise
+func ValidCloudBase(t *Transaction) (bool, CloudBaseTransactionCode) {
```
jordanschalm, line 125: This function should be aware of the block as well. It needs to verify that the output matches the block reward for the given block and it needs to verify that the output address is the same as the miner who mined the block.

blockchain/validation.go
```diff
@@ -83,13 +178,29 @@ func (bc *BlockChain) ValidBlock(b *Block) (bool, BlockCode) {
 		return false, BadBlockNumber
 	}
 
+	// Check that the first transaction is a CloudBase transaction
+	if valid, _ := ValidCloudBase(b.Transactions[0]); !valid {
```
jordanschalm, line 181: 👍

blockchain/validation.go
```diff
@@ -109,3 +220,9 @@ func (bc *BlockChain) ValidBlock(b *Block) (bool, BlockCode) {
 
 	return true, ValidBlock
 }
+
+// BigExp returns an big int pointer with the result set to base**exp,
+// if exp <= 0, the result is 1
+func BigExp(base, exp int) *big.Int {
```
jordanschalm, line 225: 👍

blockchain/validation.go
```diff
@@ -83,13 +178,29 @@ func (bc *BlockChain) ValidBlock(b *Block) (bool, BlockCode) {
...
 	}
 
+	// Check that the target is within the min and max difficulty levels
+	target := HashToBigInt(b.Target)
+	if target.Cmp(MinDifficulty) == 1 || target.Cmp(MaxDifficulty) == -1 {
+		return false, BadTarget
+	}
+
+	// Check that time is not greater than current time or equal to 0
+	if uint32(b.Time) == 0 || uint32(b.Time) > uint32(time.Now().Unix()) {
```
jordanschalm: I think this needs some more consideration. The node's clocks will not be synchronized so this will fail some of the time.

consensus/consensus.go
```diff
@@ -0,0 +1,92 @@
...
+
+const (
+	// blockRewardHalvingRate is the number of blocks that need to be mined
+	// before the blockReward is halved
+	blockRewardHalvingRate int = 210000
+)
+
+var (
+	// BlockReward is the current reward for mining a block
+	BlockReward uint64 = 25
```
jordanschalm, line 40: I think we should make this like 1024 so we all become rich!

consensus/consensus.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Check if the block is valid
+	if valid, code := bc.ValidBlock(b); !valid {
+		log.Errorf("Invalid block, BlockCode:%d", code)
+		return false, BadBlock
+	}
+
+	// Check if the CloudBase transaction reward is equal to the network's
+	// current block reward
+	if b.Transactions[0].Outputs[0].Amount != BlockReward {
```
jordanschalm: Also need to set the block reward based on the current length of the chain when the node starts up and downloads the chain

miner/miner.go
```diff
@@ -1,106 +1,75 @@
 package miner
```
jordanschalm, line 1: 👍  Awesome, so much simpler

jordanschalm (changes requested):
Looks awesome. Couple of comments

blockchain/validation.go
david-julien: any suggestions?

jordanschalm: Not really. I think it's something to consider later and for now we can just not worry about checking time.

david-julien: Ok, for now I'll just make sure its greater than 0

blockchain/validation.go
david-julien: Isn't the point of ValidCloudBase to validate the cloud base transaction before it is added to a block? The block only stores the information of the address of the recipient of the block reward in the cloudbase transaction itself. Therefore shouldnt it just take in the address and the block reward in instead of the block?

jordanschalm: Yeah it's fine to take the address and block reward rather than the block. We just need to verify those cases somehow

jordanschalm: There is also potentially a difference between validating a block that we receive from another node and validating a block that we ourselves are mining? Not sure

david-julien: Yes thats why I have the code in consensus.go that validates a received mined block and does that check with the address and the block reward to make sure they are equal to the current block reward. The miner should technically be able to set the mining values to whatever they want, as long as the output including address and reward is valid (1 output and a non 0 address and reward), but the rest of the network needs to confirm that the final mined block adheres to our consensus rules, which includes checking that the mined block has an address set to the miner that mined it, and the block reward is equal to the curent network's block reward, and the target is equal to the current network target, etc...

Issue #58, opened 2017-06-06
Push handling logic


jordanschalm:
Update internal state with the payload when a push comes in.

Issue #57, opened 2017-06-06
Request handling logic


jordanschalm:
Respond to requests in a reasonable way.

Issue #56, opened 2017-06-06
Initialization logic


jordanschalm:
Logic for getting a newly started node into a stable state.
* acquire the blockchain
* connect to some neighbouring nodes

PR #87, opened 2017-07-09
57 & 58 Request/Push Handling Logic (cont)

chadlagore:
# What Changed

+ Implemented `Request` handling logic.
+ Finished `Push` handling logic.
    + This required an update to the `miner` module, @david-julien please 👀 .

### TODO 💡 

+ `Mine` should take an interface with a callback as an argument.
+ We can use this interface as a way to update progress on mining, alert the app of progress etc etc.

```go
type MinerUpdater interface {
    UpdateProgress(i int)
    UpdateFinished(result MiningResult)
}

go miner.Mine(bc, block, updater)

// As opposed to:
result := miner.Mine(bc, block)   // => Spins forever.
```


app/app.go
```diff
@@ -159,13 +165,17 @@ func initializeWorkers() {
 	}
 }
 
+// initializeChain creates the blockchain for the node.
```
chadlagore, line 205: This is TODO.

app/app.go
```diff
@@ -61,8 +61,10 @@ func Run(cfg conf.Config) {
 	go peer.MaintainConnections()
 
 	// Request the blockchain.
-	log.Info("Requesting blockchain from peers... ")
-	RequestBlockChain()
+	if chain == nil {
```
bfbachmann, line 98: I'm doing some refactoring this week such that we don't have all these global variables, but instead we have instances of App for each Cumulus "node" we are running. The `chain` variable will also be part of that at some point just so you're aware.

app/app.go
```diff
@@ -91,19 +93,23 @@ func ConnectAndDiscover(target string) {
 // on peers whos PushHandlers have been overridden.
 func RequestHandler(req *msg.Request) msg.Response {
 	res := msg.Response{ID: req.ID}
+	typeErr := msg.NewProtocolError(msg.InvalidResourceType,
+		"Invalid resource type")
+	paramErr := msg.NewProtocolError(msg.InvalidResourceType,
```
bfbachmann, line 130: We should probably add a different error code here to be more specific since this technically isn't an invalid resource type error.

app/app.go
```diff
@@ -159,13 +165,17 @@ func initializeWorkers() {
 	}
 }
 
+// initializeChain creates the blockchain for the node.
+func initializeChain() {
+	chain, _ = blockchain.NewValidChainAndBlock()
+	// Check if chain exists on disk.
```
bfbachmann: I don't think you should check if the chain exists here because this function should only be called when we don't already have a blockchain right? Actually, either way, when a node boots it should always ask its peers for blocks anyway (regardless of whether it already has a blockchain stored locally) since there is no other way of knowing for sure that you have the most up-to-date blockchain.

app/app_test.go
```diff
@@ -13,6 +14,15 @@ func init() {
 	log.SetLevel(log.InfoLevel)
 }
 
+func createNewBlockRequest(blockNumber interface{}) *msg.Request {
```
bfbachmann, line 17: Why not `blockNumber` of type `int`?

miner/miner_test.go
```diff
@@ -27,18 +28,43 @@ func TestMine(t *testing.T) {
...
+	assert.False(t, mineResult.Complete)
+	assert.Equal(t, mineResult.Info, MiningNeverStarted)
+}
+
+func TestMineHaltMiner(t *testing.T) {
+	bc, b := blockchain.NewValidChainAndBlock()
+
+	// Set min difficulty to be equal to the target so that the block validation
+	// passes
+	blockchain.MaxTarget = new(big.Int).Sub(
```
david-julien: Reset the max target to its original value before the test returns so it doesn't cause any problems for any tests in the same file written after this test.

miner/miner.go
```diff
@@ -2,29 +2,98 @@ package miner
...
+	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/ubclaunchpad/cumulus/blockchain"
 	"github.com/ubclaunchpad/cumulus/consensus"
 )
 
+// CurrentlyMining is a flag to control the miner.
+var CurrentlyMining bool
```
david-julien: Should CurrentlyMining and the lock be private?

david-julien (approved):

app/app.go
david-julien: Don't know how we're all handling TODOs but should we mark this as TODO in the comment just to make sure we don't forget?

miner/miner.go
chadlagore: Good call. Originally I wanted to have other packages check on them, but the exported function does that now.

blockchain/blockchain.go
```diff
@@ -71,3 +71,12 @@ func (bc *BlockChain) ContainsTransaction(t *Transaction, start, stop uint32) (b
 	}
 	return false, 0, 0
 }
+
+// CopyLocalBlockByIndex returns a copy of a block in the local chain by index.
+func (bc *BlockChain) CopyBlockByIndex(i uint32) *Block {
+	blk := bc.Blocks[i]
```
bfbachmann: What if this is an out-of-bounds reference? I think you can also aggregate this line and the one below into one.

miner/miner.go
```diff
@@ -2,29 +2,98 @@ package miner
...
 	log "github.com/Sirupsen/logrus"
 	"github.com/ubclaunchpad/cumulus/blockchain"
 	"github.com/ubclaunchpad/cumulus/consensus"
 )
 
+// currentlyMining is a flag to control the miner.
+var currentlyMining bool
+
+// currentlyMiningLock is a read/write lock to change the Mining flag.
+var currentlyMiningLock sync.RWMutex
```
bfbachmann, line 17: I guess this might be overkill but you could put `currentlyMining` and `currentlyMiningLock` into a struct of type `Miner`. Then you could make the functions below receivers of the `Miner` type. That way we can manage multiple instances of `Miner`.

miner/miner.go
```diff
@@ -2,29 +2,98 @@ package miner
...
+
+const (
+	// MiningSuccessful is returned when the miner mines a block.
+	MiningSuccessful = iota
```
bfbachmann, line 21: You only need to set the first constant to `iota`, the compiler will fill in the following ones on its own (see https://golang.org/ref/spec#Constant_declarations and the section below).

miner/miner.go
```diff
@@ -2,29 +2,98 @@ package miner
...
+}
+
 // Mine continuously increases the nonce and tries to verify the proof of work
 // until the puzzle is solved
-func Mine(bc *blockchain.BlockChain, b *blockchain.Block) bool {
+// TODO: Make Mine take an interface with a callback as an arguement.
+func Mine(bc *blockchain.BlockChain, b *blockchain.Block) *MiningResult {
+	setStart()
 	if valid, _ := bc.ValidBlock(b); !valid {
 		log.Error("Invalid block")
```
bfbachmann: I know you added this earlier but we should probably add more detail here. Generally when we surface errors we want to know what package they're from and what action what being taken when the error occurred.
In this case something like "Mine failed, miner received an invalid block".

miner/miner.go
```diff
@@ -2,29 +2,98 @@ package miner
...
+	// MiningSuccessful is returned when the miner mines a block.
+	MiningSuccessful = iota
+	// MiningNeverStarted is returned when the block header is invalid.
+	MiningNeverStarted = iota
+	// MiningHalted is returned when the app halts the miner.
+	MiningHalted = iota
+)
+
+// MiningResult contains the result of the mining operation.
+type MiningResult struct {
```
bfbachmann: Why have this be a type when we can just return `(complete bool, info int)` from `Mine()`? Do we use MiningResult outside this file?

miner/miner.go
```diff
@@ -2,29 +2,98 @@ package miner
...
+}
+
+func setStart() {
+	currentlyMiningLock.Lock()
+	defer currentlyMiningLock.Unlock()
+	currentlyMining = true
+}
+
+// StopMining stops the miner from mining.
+func StopMining() {
```
bfbachmann, line 85: I love the idea of cancelling a mining job. Nice work, we'll definitely need to do this when we are notified of new blocks that were added to the chain that contain transactions we are trying to mine in this block.

bfbachmann (commented):
Nice work

miner/miner.go
bfbachmann: Edit: I see you want to make `Mine()` take an interface and a callback as a means of updating progress. I think using a `Miner` type could help with that. The only thing is, I fell like in mining there is no notion of progress since you can never know for sure how close you are to the solution.

app/app.go
chadlagore: Good stuff :+1:

miner/miner.go
chadlagore: Actually, I don't even think the miner should be worried about validation. Good eye 👀

app/app_test.go
chadlagore: 😆

jordanschalm (approved):
🚢

miner/miner.go
chadlagore: 1. As for `Miner` type, I agree. Not in scope here, but I think we should do it. It would be nice to build in the ability to scale your miners outwards instead of upwards.
2. By progress I more meant, how many hashes it has attempted. Progress = `number_attempted / number_of_hashes_to_needed_to_likely_produce_a_match`.

jordanschalm:
@chadlagore Feel free to merge this after review comments and merge conflicts are resolved 👍

jordanschalm:
closes #56 
closes #57 
closes #58 

Let's close these and make more descriptive issues next meeting

PR #88, opened 2017-07-10
85-block-extra-data

david-julien:


blockchain/block.go
```diff
@@ -28,6 +27,9 @@ type BlockHeader struct {
 	Time uint32
 	// Nonce starts at 0 and increments by 1 for every hash when mining
 	Nonce uint64
+	// ExtraData is an extra field that can be filled with arbitrary data to
+	// be stored in the block
+	ExtraData []byte
```
chadlagore, line 26: I know there is some analogue to this in bitcoin. Interested, why are we adding it to our chain?

blockchain/block.go
```diff
@@ -48,10 +50,18 @@ func (bh *BlockHeader) Marshal() []byte {
...
 	buf = append(buf, tempBufTime...)
 	buf = append(buf, tempBufNonce...)
+	buf = append(buf, bh.ExtraData...)
 
 	return buf
 }
 
+// Len returns the length in bytes of the BlockHeader.
+func (bh *BlockHeader) Len() int {
+	l := 2*(32/8) + 64/8 + 2*HashLen
```
chadlagore: `unsafe` doesn't seem ideal, but isn't there some notion of `sizeof` in golang?https://stackoverflow.com/questions/26975738/how-to-get-memory-size-of-variable

blockchain/block.go
```diff
@@ -9,10 +9,9 @@ import (
 )
 
 const (
-	// BlockSize is the maximum size of a block in bytes when marshaled (about 250K).
-	BlockSize = 1 << 18
-	// BlockHeaderLen is the length in bytes of a block header.
-	BlockHeaderLen = 2*(32/8) + 64/8 + 2*HashLen
+	// MaxBlockSize is the maximum size of a block in bytes when marshaled
+	// (about 250K).
+	MaxBlockSize = 1 << 18
```
chadlagore: Block size is busy blowing the bitcoin universe up - are we sure we want to enforce it? 
+ https://medium.com/@olivierjanss/the-solution-to-the-block-size-debate-42c3814ff481
+ https://www.bitcoinunlimited.info/

Did you guys discuss this on Saturday? I'm game if we're consciously making the decision.

blockchain/validation_test.go
```diff
@@ -442,11 +465,43 @@ func TestValidGenesisBlockNilGenesisBlock(t *testing.T) {
...
+	}
+
+	bc := &BlockChain{
+		Blocks: []*Block{gb},
+		Head:   HashSum(gb),
+	}
+
+	valid, code := bc.ValidGenesisBlock(gb)
+
+	if valid {
```
chadlagore: Use `assert` if we can? 😄

chadlagore (commented):
Interested in what the use case of the extra data is? I know there is some analogue in BTC. Also, why are we enforcing a block size now?

blockchain/block.go
david-julien: I was thinking its a standard blockchain feature, lets you store data for "eternity" and will potentially help us with future applications of our blockchain.

blockchain/block.go
david-julien: Bringing this to slack

blockchain/block.go
david-julien: been looking in to this, unsafe.Sizeof causes some issues because in the ExtraData byte array, Sizeof returns the size of the slice descriptor, not the size of the memory referenced by the slice which is what we want as we are looking for the length when marshalled.

Statically setting the length here should be safe if we have tests that will fail once someone messes with it.

blockchain/block.go
jordanschalm: Should we make this an array to indicate that it should be size-limited? (can change later too).

blockchain/validation.go
```diff
@@ -73,7 +73,7 @@ const (
 	BadGenesisBlockNumber
 	// BadGenesisTarget is returned when the genesis block's target is invalid.
 	BadGenesisTarget
-	// BadGenesisTime is returned when teh gensis block's time is invalid.
+	// BadGenesisTime is returned when the gensis block's time is invalid.
```
jordanschalm: there's another typo here lol (gensis -> genesis)

jordanschalm (approved):

blockchain/block.go
```diff
@@ -60,16 +62,12 @@ type Block struct {
 
 // Len returns the length in bytes of the Block.
 func (b *Block) Len() int {
-	l := BlockHeaderLen
-	for _, t := range b.Transactions {
-		l += t.Len()
-	}
-	return l
+	return len(b.Marshal())
```
bfbachmann, line 65: Nice

blockchain/validation.go
bfbachmann: Gensyscalls

blockchain/validation.go
```diff
@@ -190,6 +190,7 @@ func ValidCloudBase(t *Transaction) (bool, CloudBaseTransactionCode) {
 
 // ValidGenesisBlock checks whether a block is a valid genesis block.
 func (bc *BlockChain) ValidGenesisBlock(gb *Block) (bool, GenesisBlockCode) {
+
```
bfbachmann, line 193: This is my favorite line

blockchain/validation.go
```diff
@@ -234,12 +235,12 @@ func (bc *BlockChain) ValidGenesisBlock(gb *Block) (bool, GenesisBlockCode) {
...
 func (bc *BlockChain) ValidBlock(b *Block) (bool, BlockCode) {
 
-	// Check if the block is equal to nil
+	// Check if the block is equal to nil.
 	if b == nil {
 		return false, NilBlock
 	}
 
-	// Check if the block is the genesis block
+	// Check if the block is the genesis block.
```
bfbachmann, line 243: Good catch

bfbachmann (approved):
Wow! Great work!

blockchain/block.go
david-julien: This is what chad and i were talking about on slack, I think we decided to not limit any size for now. I made a ticket for it.

jordanschalm:
@david-julien Does this build for you locally? I get:
```
app/worker.go:76: work.Miner undefined (type BlockWork has no field or method Miner)
app/worker.go:76: too many arguments in call to chain.AppendBlock
	have (*blockchain.Block, <T>)
	want (*blockchain.Block)
```

Maybe why Travis is failing

Issue #92, opened 2017-07-12
Incoming Block Handling Logic


chadlagore:
## More Push Handling Logic 💥 

👉 Refresh the transaction pool when a new block comes in.
👉 Restart the mining job with a new set of transactions.

PR #94, opened 2017-07-13
Incoming Block Handling Logic

chadlagore:
**Status:** _Ready to merge_

# What Changed

- [x] Restart the mining job with a new set of transactions.
- [x] Refresh the transaction pool when a new block comes in.

https://github.com/ubclaunchpad/cumulus/issues/92

## TODO

- [ ] Create new error types for request errors.

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -6,9 +6,14 @@ import "github.com/ubclaunchpad/cumulus/blockchain"
 type User struct {
 	// Wallet holds the users wallets (currently just support 1).
 	Wallet blockchain.Wallet
+	// UserBlockSize is the maximum size of a block in bytes when marshaled
+	// as specified by the user.
+	BlockSize uint32
 }
 
 var currentUser *User
+
+const defaultBlockSize = 1 << 18
 
 func init() {
 	// Temporary to create a new user for testing.
@@ -18,7 +23,8 @@ func init() {
 // NewUser creates a new user
 func NewUser() *User {
 	return &User{
-		Wallet: blockchain.NewWallet(),
+		Wallet:    blockchain.NewWallet(),
+		BlockSize: defaultBlockSize,
 	}
 }
 
--- a/app/user.go
+++ b/app/user.go
@@ -1,0 +1,28 @@
+package app
+
+import "github.com/ubclaunchpad/cumulus/blockchain"
+
+// User holds basic user information.
+type User struct {
+	// Wallet holds the users wallets (currently just support 1).
+	Wallet blockchain.Wallet
+}
+
+var currentUser *User
+
+func init() {
+	// Temporary to create a new user for testing.
+	currentUser = NewUser()
+}
+
+// NewUser creates a new user
+func NewUser() *User {
+	return &User{
+		Wallet: blockchain.NewWallet(),
+	}
+}
+
+// GetCurrentUser returns the current user.
+func GetCurrentUser() *User {
+	return currentUser
+}
--- a/app/worker.go
+++ b/app/worker.go
@@ -72,13 +72,18 @@ func (w *AppWorker) HandleTransaction(work TransactionWork) {
 	}
 }
 
-// HandleBlock handles TransactionWork.
+// HandleBlock handles new instance of BlockWork.
 func (w *AppWorker) HandleBlock(work BlockWork) {
-	ok, _ := chain.ValidBlock(work.Block)
+	ok := tpool.Update(work.Block, chain)
+
 	if ok {
+		// Append to the chain before requesting
+		// the next block so that the block
+		// numbers make sense.
 		chain.AppendBlock(work.Block)
+		blk := tpool.NextBlock(chain)
 		if miner.IsMining() {
-			miner.RestartMiner(chain, work.Block)
+			miner.RestartMiner(chain, blk)
 		}
 	}
 
--- a/app/worker.go
+++ b/app/worker.go
@@ -77,12 +77,13 @@ func (w *AppWorker) HandleBlock(work BlockWork) {
 	validBlock := tpool.Update(work.Block, chain)
 
 	if validBlock {
+		user := GetCurrentUser()
 		// Append to the chain before requesting
 		// the next block so that the block
 		// numbers make sense.
 		chain.AppendBlock(work.Block)
-		address := GetCurrentUser().Wallet.Public()
-		blk := tpool.NextBlock(chain, address)
+		address := user.Wallet.Public()
+		blk := tpool.NextBlock(chain, address, user.BlockSize)
 		if miner.IsMining() {
 			miner.RestartMiner(chain, blk)
 		}
--- a/app/worker.go
+++ b/app/worker.go
@@ -76,12 +76,13 @@ func (w *AppWorker) HandleTransaction(work TransactionWork) {
 func (w *AppWorker) HandleBlock(work BlockWork) {
 	validBlock := tpool.Update(work.Block, chain)
 
-	if ok {
+	if validBlock {
 		// Append to the chain before requesting
 		// the next block so that the block
 		// numbers make sense.
 		chain.AppendBlock(work.Block)
-		blk := tpool.NextBlock(chain)
+		address := GetCurrentUser().Wallet.Public()
+		blk := tpool.NextBlock(chain, address)
 		if miner.IsMining() {
 			miner.RestartMiner(chain, blk)
 		}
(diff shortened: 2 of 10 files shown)


blockchain/transaction.go
```diff
@@ -5,6 +5,15 @@ import (
 	"io"
 )
 
+const (
+	// TxHashPointerLen is the length in bytes of a hash pointer.
+	TxHashPointerLen = 32/8 + HashLen
```
david-julien: This should be:
TxHashPointerLen = 2*(32/8) + HashLen

pool/pool.go
```diff
@@ -107,20 +107,51 @@ func (p *Pool) Update(b *blockchain.Block, bc *blockchain.BlockChain) bool {
...
+			LastBlock:   lastHash,
+			Time:        uint32(time.Now().Second()),
+			Nonce:       0,
+		}, Transactions: txns,
+	}
+
+	// Try to grab as many transactions as the block will allow.
+	// Test each transaction to see if we break size before adding.
+	for p.Len() > 0 {
+		if b.Len()+p.Peek().Len() < blockchain.BlockSize {
```
david-julien: I thought we weren't going to mess with a max block size for now?

pool/pool.go
```diff
@@ -107,20 +107,51 @@ func (p *Pool) Update(b *blockchain.Block, bc *blockchain.BlockChain) bool {
...
+	// Hash the last block in the chain.
+	ix := len(chain.Blocks) - 1
+	lastHash := blockchain.HashSum(chain.Blocks[ix])
+
+	// Build a new block for mining.
+	b := &blockchain.Block{
+		BlockHeader: blockchain.BlockHeader{
+			BlockNumber: uint32(len(chain.Blocks)),
+			LastBlock:   lastHash,
+			Time:        uint32(time.Now().Second()),
```
david-julien: We're using Unix Time stamp:
time.Now().Unix()

pool/pool.go
```diff
@@ -107,20 +107,51 @@ func (p *Pool) Update(b *blockchain.Block, bc *blockchain.BlockChain) bool {
...
+// NextBlock produces a new block from the pool for mining.
+func (p *Pool) NextBlock(chain *blockchain.BlockChain) *blockchain.Block {
+	var txns []*blockchain.Transaction
+
+	// Hash the last block in the chain.
+	ix := len(chain.Blocks) - 1
+	lastHash := blockchain.HashSum(chain.Blocks[ix])
+
+	// Build a new block for mining.
+	b := &blockchain.Block{
```
david-julien, line 140: Make sure we are calling miner.CloudBase on this block after adding transactions to it so that the first transaction in the list of transactions is the cloudbase transaction awarded to the miner.

blockchain/block.go
```diff
@@ -8,6 +8,15 @@ import (
 	"io"
 )
 
+const (
```
david-julien: If we're going to use these constants, we should make tests that assert their size is correct so that if any extra fields get added to the block header for example, and the developer doesn't change the size constant, it will fail.

This can be done by creating a block header, and asserting that the len(blockheader.Marshal()) is the same as the BlockHeaderLen. (Assuming that the developer appended the extra fields in the marshalling function)

It might also be a good idea to make these read only by making them private and creating "getter" functions for them.

app/worker.go
```diff
@@ -72,13 +72,18 @@ func (w *AppWorker) HandleTransaction(work TransactionWork) {
...
+// HandleBlock handles new instance of BlockWork.
 func (w *AppWorker) HandleBlock(work BlockWork) {
-	ok, _ := chain.ValidBlock(work.Block)
+	ok := tpool.Update(work.Block, chain)
+
 	if ok {
+		// Append to the chain before requesting
+		// the next block so that the block
+		// numbers make sense.
 		chain.AppendBlock(work.Block)
```
jordanschalm: It seems like `AppendBlock` doesn't do any validation, but we must validate the block before we add it to our chain.

blockchain/block.go
jordanschalm: Agreed, although they are already read-only because they are `const`s

jordanschalm (commented):
Looks good once David's concerns are addressed

blockchain/block.go
david-julien: haha sorry I was thinking they were vars at the end there for some reason...

blockchain/transaction.go
```diff
@@ -5,6 +5,15 @@ import (
 	"io"
 )
 
+const (
```
david-julien: Same goes for these constants as mentioned above^

pool/pool.go
chadlagore: True, but:

1. We need to determine how many transactions go into a single block somehow.

2. I haven't fixed anything. The `BlockSize` may be configurable in the future via the console or by gas or whatever. For the time being this is MVP that is easily adjusted by playing with the `BlockSize` variable (and making it `var`).

Do we have any other ideas about how to make a block at the moment?

blockchain/block.go
chadlagore: I'll just say I didn't add these variables on this PR - I just added `MaxTransactionsPerBlock`, not sure why it's all green 🤔. And I totally agree about the constant size.

app/worker.go
chadlagore: For sure. We validate it above with `pool.Update` - I'll add comments.

blockchain/transaction.go
chadlagore: Why do we even have these. We should just enforce `len(thing.Marshal())`.

blockchain/block.go
david-julien: hmm thats strange haha, I was wondering why they were added in again. I think you maybe pulled before https://github.com/ubclaunchpad/cumulus/pull/88 was merged because I got rid of all of these.

chadlagore: Ahh, I'll rebase. Problem solved.

blockchain/transaction.go
david-julien: I enforced this in https://github.com/ubclaunchpad/cumulus/pull/88 as well! will get fixed once you rebase.

chadlagore: Weeeeeeee

pool/pool.go
david-julien: you can also merge https://github.com/ubclaunchpad/cumulus/pull/95 and use util.UnixNow(). Wrapper function for uint32(time.Now().Unix()).

app/user.go
```diff
@@ -0,0 +1,34 @@
...
+	return &User{
+		Wallet:    blockchain.NewWallet(),
+		BlockSize: defaultBlockSize,
+	}
+}
+
+// GetCurrentUser returns the current user.
+func GetCurrentUser() *User {
+	return currentUser
+}
```
bfbachmann, line 34: Don't need this

app/worker.go
```diff
@@ -62,30 +62,37 @@ func (w AppWorker) Start() {
...
+
+	if validBlock {
+		user := GetCurrentUser()
+		// Append to the chain before requesting
+		// the next block so that the block
+		// numbers make sense.
 		chain.AppendBlock(work.Block)
+		address := user.Wallet.Public()
+		blk := tpool.NextBlock(chain, address, user.BlockSize)
 		if miner.IsMining() {
```
bfbachmann, line 87: I think we should check if the transactions in the new block collide with the block we are currently hashing. Only if there are collisions we should restart the miner. Otherwise restarting it would be a waste.

pool/pool.go
```diff
@@ -107,20 +109,56 @@ func (p *Pool) Update(b *blockchain.Block, bc *blockchain.BlockChain) bool {
...
-	txns := make([]*blockchain.Transaction, l)
-	for i := 0; i < l; i++ {
-		t := p.GetN(i)
-		txns[i] = t
-		p.Delete(t)
+	return nil
+}
+
+// NextBlock produces a new block from the pool for mining.
+func (p *Pool) NextBlock(chain *blockchain.BlockChain,
```
bfbachmann, line 131: What do you think about calling it something like `MakeBlock` or `NewBlock`. I feel like `NextBlock

bfbachmann (approved):
Nice

app/worker.go
jordanschalm: @bfbachmann We need to restart it regardless because the block number will be different

PR #95, opened 2017-07-13
93 create common folder

david-julien:


jordanschalm (changes requested):
Really like this! I think having subpackages (math) under `common` is overkill. I'd prefer either all in one package or just two packages `util` for utility functions and `constants` for the constants.

PR #97, opened 2017-07-13
96 create pr template

david-julien:


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
+## Status
+**READY/IN DEVELOPMENT**
+
+## Related Issue
```
jordanschalm, line 4: ZenHub has a feature for doing this at the bottom of the PR page. Let's use that instead (we could put a note in the template to connect the PR to the issue using that feature)

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+[link]()
+
+## Description
+* A few sentences describing the overall goals of the pull request's commits.
+* List general components of the application that this PR will affect.
+
+## WIKI Updates
+* [link]()
```
jordanschalm, line 12: Let's just put one line of links here so we don't have to delete a bunch of lines for each PR

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+* ...
+
+## Todos
+
+General:
+- [ ] Tests
+- [ ] Documentation
+- [ ] Wiki
+
+Other (links to TODOs in code):
```
jordanschalm, line 23: Same, don't think we need all these links as we'll be deleting them each time

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+* [link]()
+* [link]()
+* ...
+
+## Extra Changes/Bug Fixes
+List any extra quick changes or bug fixes on this branch.
+
+* ...
```
jordanschalm, line 31: Ditch this

jordanschalm (changes requested):
I think we should make the description text that the PR creator will have to replace with their own text as short and easy to replace as possible (one line each). Otherwise looks good and I think it'll help a lot

PR #103, opened 2017-07-16
90 app peer refactor

### LINKED ISSUE #90: Refactor App and Peer so they don't expose unnecessary globals
Currently the App and Peer packages expose a number of global variables that need not be exposed. 
We should refactor these packages to use struct types that contain these variables and change their exported functions to be receivers of their respective struct types. The code will be much cleaner and easier to understand that way.

bfbachmann:
## Related Issue
Closes #90 

## Description
* Refactors the App and Peer packages so they don't unnecessarily expose global variables.
* Adds WaitGroup to critical functions that are run as goroutines (like Listen and MaintainConnections) so we can guarantee that they are running before opening the console or sending requests to them in tests.
* Moves Peer and PeerStore code into two different files

## WIKI Updates
* None. I didn't change any functionality.

## Todos

General:
- [ ] Write integration tests using this refactored code


peer/peerstore.go
```diff
@@ -0,0 +1,153 @@
+package peer
```
bfbachmann, line 1: This whole file contains all the code that used to be in peer.go that involved the PeerStore. No need to review it in detail, I didn't change any of the functionality.

app/app.go
```diff
@@ -25,12 +26,22 @@ var (
 	tpool *pool.Pool
```
bfbachmann, line 25: Left these private package-level variables here because they should not be instantiable as part of App. They should only exist in one place in the program.

jordanschalm: This definitely should not be global. The set of transactions a node/app instance knows about should not be shared by another instance just because they happen to be running in the same process. Pretty sure this should be a field on the app struct.

app/app.go
```diff
@@ -25,12 +26,22 @@ var (
...
+	PeerStore *peer.PeerStore
+}
+
 // Run sets up and starts a new Cumulus node with the
-// given configuration.
-func Run(cfg conf.Config) {
+// given configuration. This should only be called once (except in tests)
+func Run(cfg conf.Config, wg *sync.WaitGroup) {
 	log.Info("Starting Cumulus node")
 	config = &cfg
```
jordanschalm: config should be part of the app struct, not be global

cmd/run.go
```diff
@@ -26,7 +28,17 @@ var runCmd = &cobra.Command{
 			Verbose:   verbose,
 			Console:   console,
 		}
-		app.Run(config)
+
+		// Set up wait group to signal when we have started listening and
+		// maintaining connections
+		wg := &sync.WaitGroup{}
```
jordanschalm: Why not just instantiate this in `app.Run`?

conn/conn_test.go
```diff
@@ -33,7 +33,7 @@ func TestListen(t *testing.T) {
 		wg.Done()
 	}
 
-	go Listen(":8080", handler)
+	go Listen(":8080", handler, &wg)
 	// Sleep to guarantee that our listener is ready when we start making connections
 	time.Sleep(time.Millisecond)
```
jordanschalm: No longer necessary, we have a waitgroup now!

jordanschalm (changes requested):
A couple of objects that need to be moved into app still, besides that awesome refactor

app/app.go
jordanschalm: ^^

app/app.go
```diff
@@ -57,73 +72,72 @@ func Run(cfg conf.Config) {
...
 
-	// Return to command line.
-	select {} // Hang main thread. Everything happens in goroutines from here
+	// Wait for goroutines to start
+	wg.Wait()
 }
 
 // ConnectAndDiscover tries to connect to a target and discover its peers.
-func ConnectAndDiscover(target string) {
+func (a App) ConnectAndDiscover(target string) {
```
jordanschalm: Let's have the received be an app pointer to avoid copying. Not a big deal but standard practice

app/console.go
```diff
@@ -2,19 +2,29 @@ package app
...
 	"github.com/abiosoft/ishell"
 	"github.com/ubclaunchpad/cumulus/blockchain"
-	"github.com/ubclaunchpad/cumulus/conn"
 	"github.com/ubclaunchpad/cumulus/peer"
 )
 
-var shell *ishell.Shell
+var (
+	shell     *ishell.Shell
+	peerStore *peer.PeerStore
```
jordanschalm, line 13: Don't love that these are still global but I guess it won't matter

jordanschalm (approved):

app/app.go
bfbachmann: I can't put this into app because it has dependencies in a bunch of other places. So I say we merge, wait for the App refactor on Chad's end, and then pull the variables into the App type. Either that or we merge this and Chad does it as part of his App refactor PR.

app/app.go
```diff
@@ -18,24 +19,38 @@ import (
...
 var (
-	config  *conf.Config
 	chain   *blockchain.BlockChain
 	logFile = os.Stdout
 	// A reference to the transaction pool
 	tpool *pool.Pool
 )
 
+// App contains information about a running instance of a Cumulus node
+type App struct {
```
chadlagore, line 29: make private?

app/console.go
chadlagore: Agreed, I see what you're doing, but it is a strange place to have these globals floating around.

app/app.go
bfbachmann: Yea, I guess we could. It's not really used in other packages anyway. What do you think @jordanschalm?

chadlagore (approved):
Nothing blocking :+1:

app/console.go
bfbachmann: Yea I'm not really sure what to do about that. I guess I could make my own Shell type that is composed of ishell.Shell and the Peerstore. The thing is, there wouldn't really be much point because that Shell type wouldn't be used anywhere else and would only ever be instantiated once throughout the lifetime of the program.

PR #105, opened 2017-07-22
100 msg to json

bfbachmann:
## Related Issue
#100 

## Description
* This change moves our message encoding and decoding to JSON. Specifically it changes messages so there is now a concrete `Message` type (rather than the interface we used before) that holds the payload type and the payload in byte slice form. This allows us to correctly decode the payload when a message is received. 
* Payloads can be one of `Push`, `Request`, and `Response`. 
* The `Read` function will return a `MessagePayload` (essentially just an interface that wraps the Request, Response, and Push concrete types). 
* Callers of `Read` can switch on the returned payload type to retrieve its underlying concrete type.

## WIKI Updates
* [Peer-to-Peer](https://github.com/ubclaunchpad/cumulus/wiki/Peer-to-Peer)

## Todos
None. Cuz i get werk done.

### DIFF
--- a/msg/message.go
+++ b/msg/message.go
@@ -1,10 +1,11 @@
 package msg
 
 import (
-	"encoding/gob"
+	"encoding/json"
+	"fmt"
 	"io"
 
-	"github.com/ubclaunchpad/cumulus/blockchain"
+	log "github.com/Sirupsen/logrus"
 )
 
 type (
@@ -37,19 +38,10 @@ const (
 	SubnetFull = 503
 )
 
-// Initializes all the types we need to encode.
-func init() {
-	gob.Register(&Request{})
-	gob.Register(&Response{})
-	gob.Register(&Push{})
-	gob.Register(&blockchain.Transaction{})
-	gob.Register(&blockchain.Block{})
-}
-
 // ProtocolError is an error that occured during a request.
 type ProtocolError struct {
-	Code    ErrorCode
-	Message string
+	Code    ErrorCode `json:"Code"`
+	Message string    `json:"Message"`
 }
 
 // NewProtocolError returns a new error struct.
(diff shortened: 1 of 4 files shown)


msg/message.go
```diff
@@ -32,24 +33,12 @@ const (
...
-	gob.Register(&Push{})
-	gob.Register(&blockchain.Transaction{})
-	gob.Register(&blockchain.Block{})
-}
-
 // ProtocolError is an error that occured during a request.
 type ProtocolError struct {
-	Code    ErrorCode
-	Message string
+	Code    ErrorCode `json:"Code"`
```
jordanschalm: If the json struct tag is the same as the struct field name, then it can be omitted. I'd say either remove the struct tags or make them `camelCase` like typical JSON

msg/message.go
```diff
@@ -73,49 +72,102 @@ type Message interface {
 // parameters. PeerInfo requests should send all info of all peers. Block requests
 // should specify block number in parameters.
 type Request struct {
-	ID           string
-	ResourceType ResourceType
-	Params       map[string]interface{}
+	ID           string                 `json:"ID"`
```
jordanschalm: same point applies in these structs for json tags

peer/peer.go
```diff
@@ -223,17 +223,18 @@ func (p *Peer) Dispatch() {
 		}
 		errCount = 0
 
-		switch message.(type) {
+		switch payload.(type) {
```
jordanschalm: Nice, I like how you avoided needing changes outside `message` package

peer/peer.go
```diff
@@ -316,15 +315,19 @@ func MaintainConnections() {
 // PeerInfoHandler will handle the response to a PeerInfo request by attempting
 // to establish connections with all new peers in the given response Resource.
 func PeerInfoHandler(res *msg.Response) {
-	peers := res.Resource.([]string)
+	peers := res.Resource.([]interface{})
```
jordanschalm, line 318: Why is this necessary?

peer/peer.go
```diff
@@ -370,24 +373,28 @@ func exchangeListenAddrs(c net.Conn, d time.Duration) (string, error) {
 		receivedAddr := false
 		sentAddr := false
 		var addr string
+		var ok bool
 
 		for !receivedAddr || !sentAddr {
-			message, err := msg.Read(c)
+			payload, err := msg.Read(c)
```
jordanschalm: Minor thing but I think calling this `message` still makes sense. Technically we're extracting the "payload" from the "message" in the message package but from the context of peer this is a message.

jordanschalm (approved):
Couple minor suggestions

peer/peer.go
bfbachmann: This is because when arrays are unmarshalled from JSON they become `[]interface{}` in Go.

jordanschalm: Right but we're type casting them. Why can't we type cast it directly to `[]string` like before rather than type casting each element individually?

Issue #102, opened 2017-07-15
Generate public/private key-pair from console


jordanschalm:
# Console Options

👉 Get the user set up with a `Wallet`.

PR #108, opened 2017-07-23
102 generate private public key from console

chadlagore:
## Related Issue
[102](https://github.com/ubclaunchpad/cumulus/issues/102)

## Description
* Added the ability to generate a shortened public address from the console.
* I started by using the Blockchain shortening algorithm (https://en.bitcoin.it/wiki/Technical_background_of_version_1_Bitcoin_addresses), but it seems unnecessarily complex. I switched to the Ethereum approach.
* The emoji idea is a bit more complex, and I want to do it right - so next PR.

```bash
What would you like to create?
 ❯ Wallet
   Transaction
Enter wallet name: ched
💳  New hot wallet created!
🙋  Name: ched
📫  Address: 20b58af1ae1f7640b381a9e62ab9c894dc668685
>>>
```

## WIKI Updates
* [Using-the-Console](https://github.com/ubclaunchpad/cumulus/wiki/Using-the-Console)

## Todos
+ Add emoji address.

General:
- [x] Tests - minus console tests 🤔 
- [x] Documentation (in code)
- [x] Wiki

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,14 +10,14 @@ cumulus:
 test:
 	go test $(PACKAGES) --cover
 
-run-console:
-	make clean && make cumulus && ./cumulus run -c
+run-console: cumulus
+	./cumulus run -c
 
 deps:
 	glide install
 
 clean:
-	rm cumulus || true
+	rm cumulus
 
 install-glide:
 	sh scripts/install_glide.sh
--- a/Makefile
+++ b/Makefile
@@ -10,11 +10,14 @@ cumulus:
 test:
 	go test $(PACKAGES) --cover
 
+run-console:
+	make clean && make cumulus && ./cumulus run -c
+
 deps:
 	glide install
 
 clean:
-	rm cumulus
+	rm cumulus || true
 
 install-glide:
 	sh scripts/install_glide.sh
--- a/app/console.go
+++ b/app/console.go
@@ -6,6 +6,7 @@ import (
 	"github.com/abiosoft/ishell"
 	"github.com/ubclaunchpad/cumulus/blockchain"
 	"github.com/ubclaunchpad/cumulus/peer"
+	"gopkg.in/kyokomi/emoji.v1"
 )
 
 var (
@@ -48,7 +49,7 @@ func RunConsole(ps *peer.PeerStore) {
 	})
 
 	shell.Start()
-	shell.Println("Cumulus Console")
+	emoji.Println(":cloud: Welcome to the :sunny: Cumulus console :cloud:")
 }
 
 func create(ctx *ishell.Context) {
@@ -57,7 +58,7 @@ func create(ctx *ishell.Context) {
 		"Transaction",
 	}, "What would you like to create?")
 	if choice == 0 {
-		ctx.Println("New Wallet:", blockchain.NewWallet().Public())
+		createHotWallet(ctx)
 	} else {
 		shell.Print("Sender wallet ID: ")
 		senderID := shell.ReadLine()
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -10,11 +10,14 @@ cumulus:
 test:
 	go test $(PACKAGES) --cover
 
+run-console:
```
jordanschalm: You can specify a dependency for `run-console` like
```
run-console: cumulus
```
so that cumulus must be up-to-date before we run the `run-console` target.

Might make this code cleaner

Makefile
```diff
@@ -10,11 +10,14 @@ cumulus:
...
+	make clean && make cumulus && ./cumulus run -c
+
 deps:
 	glide install
 
 clean:
-	rm cumulus
+	rm cumulus || true
```
jordanschalm: Above would allow removing this

app/console.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/ubclaunchpad/cumulus/blockchain"
 	"github.com/ubclaunchpad/cumulus/conn"
 	"github.com/ubclaunchpad/cumulus/peer"
+	"gopkg.in/kyokomi/emoji.v1"
```
jordanschalm: 🙌

app/user.go
```diff
@@ -4,13 +4,16 @@ import "github.com/ubclaunchpad/cumulus/blockchain"
...
-	// Wallet holds the users wallets (currently just support 1).
-	Wallet blockchain.Wallet
-	// UserBlockSize is the maximum size of a block in bytes when marshaled
-	// as specified by the user.
+	HotWallet
 	BlockSize uint32
 }
 
+// HotWallet is a representation of the users wallet.
+type HotWallet struct {
```
jordanschalm, line 12: What's hot about it?

app/user.go
```diff
@@ -4,13 +4,16 @@ import "github.com/ubclaunchpad/cumulus/blockchain"
 
 // User holds basic user information.
 type User struct {
-	// Wallet holds the users wallets (currently just support 1).
-	Wallet blockchain.Wallet
-	// UserBlockSize is the maximum size of a block in bytes when marshaled
-	// as specified by the user.
+	HotWallet
```
jordanschalm, line 7: Is the plan for this to be embedded in the app struct? The miner at least will need to have access to this to sign mined blocks

jordanschalm (approved):
Couple minor things

app/user.go
chadlagore: Practically speaking, it's in memory.

https://en.bitcoin.it/wiki/Hot_wallet

At the moment it's a wrapper to get a name on it the wallet.

app/user.go
chadlagore: We should probably embed the current user in the app struct. The miner doesn't actually appear to sign the mined block at the moment. 🤔

Makefile
```diff
@@ -10,6 +10,9 @@ cumulus:
 test:
 	go test $(PACKAGES) --cover
 
+run-console: cumulus
```
bfbachmann, line 13: This seems like a very narrow use case for a make command. In production, most people will need to specify flags (like interface, port, etc.) which you can't do just by adding the flags to your make command. I also don't think its particularly useful because it's actually the same length as the command it wraps, and it assumes that there is a cumulus binary in the current working directory.

app/user.go
```diff
@@ -23,7 +26,10 @@ func init() {
 // NewUser creates a new user
 func NewUser() *User {
 	return &User{
-		Wallet:    blockchain.NewWallet(),
+		HotWallet: HotWallet{
+			Wallet: blockchain.NewWallet(),
```
bfbachmann, line 30: Does this work? It seems like it shouldn't because HotWallet has no member "Wallet".

blockchain/wallet.go
```diff
@@ -33,6 +37,28 @@ type Address struct {
 	X, Y *big.Int
 }
 
+// Repr returns a string representation of the address. We follow
+// ethereums protocol, replacing Keccak hash with SHA256.
+// Where pr is the private key,
+//    	A(pr) = SHA256(ECDSAPUBLICKEY(pr))[96:255],
+// Resources:
+// http://gavwood.com/paper.pdf (fig 213)
+func (a Address) Repr() string {
```
bfbachmann, line 46: Use pointer receivers so we don't copy the address.

bfbachmann (commented):
Looks good so far. Just a few minor comments.

Makefile
chadlagore: It builds the binary if it doesn't exist.

app/user.go
chadlagore: Indeed it does, `blockchain.Wallet`.

bfbachmann: You learn something every day!

Issue #106, opened 2017-07-22
Refactor work structure 


chadlagore:
Use only one worker, updating tests accordingly. Remove code that is for multiple workers.

PR #109, opened 2017-07-24
Initial Refactor to Remove the Worker Abstraction

chadlagore:
**Status**: Ready to merge.

## Related Issue
[106](https://github.com/ubclaunchpad/cumulus/issues/106)

## Description
* We've decided to :gun: the worker abstraction.

## Todos
- [x] Fix everything I just fucked up.

General:
- [ ] Tests (still working)


app/app_test.go
```diff
@@ -2,167 +2,97 @@ package app
...
+		assert.True(t, ok)
+		assert.Equal(t, tr, txn)
 	}
-	// Add more here...
 }
 
-func TestRequestHandlerNewBlockOK(t *testing.T) {
-	initializeChain()
-	a := App{PeerStore: peer.NewPeerStore("127.0.0.1:8000")}
+// TODO: Enable once block request by hash implemented.
```
chadlagore, line 49: @david-julien is working on something to fix this test

app/app.go
```diff
@@ -13,23 +13,38 @@ import (
...
+	transactionQueueSize = 100
 )
 
 // App contains information about a running instance of a Cumulus node
 type App struct {
-	PeerStore *peer.PeerStore
+	CurrentUser *User
+	PeerStore   *peer.PeerStore
+	Chain       *blockchain.BlockChain
+	Pool        *pool.Pool
```
bfbachmann, line 39: AEEE LEEETS GOOOOO!!!!

app/app.go
```diff
@@ -13,23 +13,38 @@ import (
...
 }
 
+// BlockWorkQueue is a queue of blocks to process.
+var blockQueue = make(chan *blockchain.Block, blockQueueSize)
```
bfbachmann: I would wrap these all up like this:
```
var (
    blockQueue = make(chan *blockchain.Block, blockQueueSize)
    ...
)
```

app/app.go
```diff
@@ -110,7 +121,7 @@ func Run(cfg conf.Config) {
 		}
 		log.Info("Redirecting logs to logfile")
 		log.SetOutput(logFile)
-		go RunConsole(a.PeerStore)
+		go RunConsole(a.PeerStore, &a)
```
bfbachmann: Might as well change `RunConsole` to only take the app instance as an argument since we can access the PeerStore from `a`.

app/app.go
```diff
@@ -197,42 +209,59 @@ func (a *App) PushHandler(push *msg.Push) {
...
 }
 
-// initializeChain creates the blockchain for the node.
-func initializeChain() {
-	chain, _ = blockchain.NewValidTestChainAndBlock()
-	// TODO: Check if chain exists on disk.
-	// TODO: If not, request chain from peers.
+// HandleTransaction handles new instance of TransactionWork.
+func (a *App) HandleTransaction(txn *blockchain.Transaction) {
+	validTransaction := a.Pool.Set(txn, a.Chain)
```
bfbachmann: Can we rename `Set` to `Insert` or `Add`? I feel like `Set` doesn't really make sense give what it actually does.

app/console.go
```diff
@@ -35,30 +37,36 @@ func RunConsole(ps *peer.PeerStore) {
...
-		Func: listenAddr,
+		Func: func(ctx *ishell.Context) {
+			listenAddr(ctx, peerStore)
+		},
 	})
 	shell.AddCmd(&ishell.Cmd{
 		Name: "peers",
 		Help: "show the peers this host is connected to",
-		Func: peers,
+		Func: func(ctx *ishell.Context) {
```
bfbachmann, line 45: Why wrap function calls in new anonymous functions? Couldn't you, for example, make `peers` take `ctx` and then just do the stuff `peers` currently does? That would make the code shorter and more readable.

bfbachmann: I see now that you're doing it because you only have access to `app` from `Run()`, but that need not be the case in my opinion.

bfbachmann (commented):
I like what I see Chad.. I like what I see

pool/pool.go
```diff
@@ -64,11 +71,13 @@ func getIndex(a []*PooledTransaction, target time.Time, low, high int) int {
 // Set inserts a transaction into the pool, returning
 // true if the Transaction was inserted (was valid).
 func (p *Pool) Set(t *blockchain.Transaction, bc *blockchain.BlockChain) bool {
-	if ok, _ := bc.ValidTransaction(t); ok {
+	if ok, err := bc.ValidTransaction(t); ok {
 		p.set(t)
```
jordanschalm: Will this succeed (ie add the transaction to the pool) if the transaction is valid wrt the blockchain but is not valid wrt some transaction already in the pool? ie. the input already exists in the pool. Does `p.set` handle that?

jordanschalm (approved):
Nice refactor

app/app.go
chadlagore: Sure! `Push` seems good to me. It goes on the back of the queue.

app/console.go
chadlagore: The original situation was making `app` and `peerStore` global, which seemed less than ideal to me.

pool/pool.go
chadlagore: Since the transactions are mapped by input hash, this type of situation currently would overwrite the existing transaction in the pool. Thats better than letting a double spend, but its not great. I'll work on it :+1:

app/app.go
```diff
@@ -38,7 +53,10 @@ func Run(cfg conf.Config) {
 
 	addr := fmt.Sprintf("%s:%d", config.Interface, config.Port)
 	a := App{
-		PeerStore: peer.NewPeerStore(addr),
+		PeerStore:   peer.NewPeerStore(addr),
+		CurrentUser: getCurrentUser(),
+		Chain:       getLocalChain(),
```
chadlagore, line 52: This is not great. In the case when the chain is not yet local, we don't exactly want to kick off a torrenting at this exact moment. We might want the chain download to be voluntary. A CLI request 🤔

PR #112, opened 2017-07-26
110 refactor consensus and validation

david-julien:
## Related Issue
[110](https://github.com/ubclaunchpad/cumulus/issues/110)

## Description
* Write a few sentences describing the overall goal of this PR.
    * Refactoring the validation code.
    * Merged validMinedBlock and validBlock together.
* List the general components of the application that this PR will affect.
    * Validation file in block chain was integrated into the consensus package.
    * Errors from validation were moved into their own file in the consensus package.
    * CurrentDifficulty and CurrentBlockReward were moved into their own file in the consensus package.

## WIKI Updates
* [Consensus](https://github.com/ubclaunchpad/cumulus/wiki/Consensus)

## Todos

General:
- [ ] Tests
- [ ] Documentation
- [ ] Wiki

Other (links to TODOs in code):
* [1st TODO for v2](https://github.com/ubclaunchpad/cumulus/pull/112/files#diff-3a5c8ea459d421751fb70343be214598R135)
* [2nd TODO for v2 - 2](https://github.com/ubclaunchpad/cumulus/pull/112/files#diff-3a5c8ea459d421751fb70343be214598R199)


common/constants/big.go
```diff
@@ -12,12 +12,25 @@ var (
...
 	Big1 = big.NewInt(1)
+	// Big2Exp232 is 2^232 respresented as type "math/big"
+	Big2Exp232 = util.BigExp(2, 232)
 	// Big2Exp256 is 2^256 respresented as type "math/big"
 	Big2Exp256 = util.BigExp(2, 256)
 )
 
 // Commonly used max values represented as type "math/big"
 var (
+	// MaxUint256 is the maximum uint232 number represented as type "math/big"
```
jordanschalm, line 23: MaxUint256 should be 232?

consensus/consensus.go
```diff
@@ -1,103 +1,237 @@
...
 
-var (
-	// BlockReward is the current reward for mining a block
-	BlockReward uint64 = 25
-	// CurrentDifficulty is the current hashing difficulty of the network
-	CurrentDifficulty = blockchain.MinTarget
-)
+	// Find the transaction input in the chain (by hash)
+	var input *blockchain.Transaction
+	input = bc.GetInputTransaction(t)
```
jordanschalm: can do `input := ...`

jordanschalm (approved):
Couple minor things. Looks good!

consensus/consensus.go
```diff
@@ -1,103 +1,237 @@
...
+	// compared to the target calculated for that specific block.
+	target := blockchain.HashToBigInt(gb.Target)
+	if target.Cmp(c.MaxTarget) == 1 ||
+		target.Cmp(c.MinTarget) == -1 ||
+		target.Cmp(blockchain.HashToBigInt(CurrentTarget())) != 0 {
+		return false, BadGenesisTarget
+	}
+
+	// Check that time is not greater than current time or equal to 0.
+	if uint32(gb.Time) == 0 {
```
bfbachmann: Seems like you're just checking if it's 0 here, not whether it's greater than the current time.

consensus/consensus.go
```diff
@@ -1,103 +1,237 @@
...
+	// Check if the block is the genesis block.
+	if b.BlockHeader.BlockNumber == 0 || bc.Blocks[0] == b {
+		if valid, code := VerifyGenesisBlock(bc, b); !valid {
+			log.Errorf("Invalid GenesisBlock, GenesisBlockCode: %d", code)
+			return false, BadGenesisBlock
+		}
+		return true, ValidBlock
+	}
+
+	// Check that block number is between 0 and max blocks.
```
bfbachmann: I would change this to: "Check that block number is valid" or something similar because "max blocks" isn't really correct terminology here.

consensus/consensus.go
```diff
@@ -1,103 +1,237 @@
...
+	// cumulus. Once the difficulty is dynamic, the target would need to be
+	// compared to the target calculated for that specific block.
+	target := blockchain.HashToBigInt(b.Target)
+	if target.Cmp(c.MaxTarget) == 1 ||
+		target.Cmp(c.MinTarget) == -1 ||
+		target.Cmp(blockchain.HashToBigInt(CurrentTarget())) != 0 {
 		return false, BadTarget
 	}
 
+	// Check that time is not greater than current time or equal to 0
```
bfbachmann: Check current time here too.

consensus/current.go
```diff
@@ -0,0 +1,40 @@
+package consensus
```
bfbachmann, line 1: I like this.

consensus/errors.go
```diff
@@ -0,0 +1,106 @@
+package consensus
```
bfbachmann, line 1: I like this too.

miner/miner.go
```diff
@@ -42,13 +41,6 @@ func RestartMiner(bc *blockchain.BlockChain, b *blockchain.Block) {
 // TODO: Make Mine take an interface with a callback as an arguement.
 func Mine(bc *blockchain.BlockChain, b *blockchain.Block) *MiningResult {
 	setStart()
-	if valid, _ := bc.ValidBlock(b); !valid {
```
bfbachmann, line 45: Why'd you remove this? Are only valid blocks passed to this function?

bfbachmann (approved):
Great job 🏅

consensus/consensus.go
```diff
@@ -1,103 +1,237 @@
...
+	// Check that the first transaction is a CloudBase transaction
+	if valid, code := VerifyCloudBase(bc, b.GetCloudBaseTransaction()); !valid {
+		log.Errorf("Invalid CloudBase, CloudBaseTransactionCode: %d", code)
+		return false, BadCloudBaseTransaction
+	}
+
+	// Verify every Transaction in the block.
+	for _, t := range b.Transactions[1:] {
+		if valid, code := VerifyTransaction(bc, t); !valid {
+			log.Errorf("Invalid Transaction, TransactionCode: %d", code)
```
chadlagore, line 192: If we want to log every bad transaction, thats cool but we could probably dry it up a bit by putting these logs at the caller.

miner/miner.go
chadlagore: The miner receives blocks from the `pool`, and the `app` orchestrates. No need to re-validate 👍

miner/miner_test.go
```diff
@@ -75,11 +67,11 @@ func TestCloudBase(t *testing.T) {
 
 	CloudBase(b, bc, w.Public())
```
chadlagore, line 68: Its still difficult for me to appreciate that this function is mutating the block and adding a new transaction to it. Can we rename it sometime? 🙏

consensus/consensus.go
```diff
@@ -1,103 +1,237 @@
...
+	// Verify signature of t
+	hash := blockchain.HashSum(t.TxBody)
+	if !ecdsa.Verify(t.Sender.Key(), hash.Marshal(), t.Sig.R, t.Sig.S) {
+		return false, BadSig
 	}
+
+	// Test if identical transaction already exists in chain.
+	end := uint32(len(bc.Blocks))
+	start := t.Input.BlockNumber
+	if exists, _, _ := bc.ContainsTransaction(t, start, end); exists {
```
chadlagore, line 41: I still worry that we have a hole here, but that is out of scope.

See https://github.com/ubclaunchpad/cumulus/issues/113.

common/constants/big.go
```diff
@@ -12,12 +12,25 @@ var (
 	Big0 = big.NewInt(0)
 	// Big1 is 1 represented as type "math/big"
 	Big1 = big.NewInt(1)
+	// Big2Exp232 is 2^232 respresented as type "math/big"
```
chadlagore, line 15: I see how these are slightly convenient, but I probably will avoid using them because they obfuscate the code. 

`b := big.NewInt(1) + util.BigExp(2, 232)` 👈 pretty clear what I'm doing
`b := Big1 + Big2Exp232` 🤔

consensus/consensus.go
```diff
@@ -1,103 +1,237 @@
...
-	BadTarget MinedBlockCode = iota
-	// BadBlockReward is returned when the block's rewards is not equal to the
-	// current reward of the network
-	BadBlockReward MinedBlockCode = iota
-	// BadCloudBase is returned when the address of the recipient of the
-	// CloudBase transaction is not valid
-	BadCloudBase MinedBlockCode = iota
-)
+// VerifyTransaction tests whether a transaction valid.
+func VerifyTransaction(bc *blockchain.BlockChain,
```
chadlagore, line 13: This was my silly formatting using PEP8 on my go code from before 🤦‍♂️ . Feel free to undo it.

consensus/consensus.go
```diff
@@ -1,103 +1,237 @@
...
+		}
+	}
+
+	// Check that the target is within the min and max difficulty levels and
+	// that the target is correct.
+	// TODO: CurrentTarget() is used because the difficulty is static for v1 of
+	// cumulus. Once the difficulty is dynamic, the target would need to be
+	// compared to the target calculated for that specific block.
+	target := blockchain.HashToBigInt(b.Target)
+	if target.Cmp(c.MaxTarget) == 1 ||
```
chadlagore: Doesn't this work:

```
if target.Cmp(blockchain.HashToBigInt(CurrentTarget())) != 0 {
    return false, BadTarget
}
```

Why would the `CurrentTarget()` return something out of range?

consensus/consensus.go
david-julien: there's some more complex time checking stuff that @jordanschalm and I were talking about so we just stuck with this simple check for now. I should probably add a TODO though.

chadlagore (commented):
Lookin cool 🕶

consensus/consensus.go
david-julien: same as above^

miner/miner_test.go
david-julien: hahaha sure I thought I was being clever but it is sort of hard to understand isn't it 😂

common/constants/big.go
david-julien: Its not just for clarity but it also helps us to avoid allocating new memory for the same values over and over again throughout the code as they are pointers.

consensus/consensus.go
david-julien: good call! These range checks are code from before the re-factor that I forgot to get rid of. oops 😅

consensus/consensus.go
bfbachmann: Yea I was worried that we are only checking the blockchain from the input transaction to the current block, but I guess that makes sense because the earlier of the two double spend transactions could not have come before it's own input transaction. Am I right there?

consensus/consensus.go
david-julien: Haha no I like it! Trying to keep the line length under 80 chars

consensus/current.go
david-julien: I'm glad you like this bruno 😗😗

consensus/consensus.go
chadlagore: All this `ContainsTransaction` bit does is check if any transaction in the chain hashes to `HashSum(t)`. It literally asserts `HashSum(t) != HashSum(t0)` for all intervening transactions. 

If Alice has 10 and sends Bob 10 then sends Bob 9.99, I don't think we catch it because the outputs look different and that changes the hashes. We may catch it elsewhere, but I can't see it 🤔

david-julien: I didn't write this haha so I wasn't trying to fix anything from 113. I think imma leave it as is, as its a re-factor and we'll handle the errors from 113 on another PR.

chadlagore: Yeah definitely out of scope, ignore us @david-julien 😄

consensus/consensus.go
chadlagore: I know the urge 😅

chadlagore:
Sorry for the merge conflict, happy to see the validation finally move out of the `blockchain` package!

common/constants/big.go
david-julien: yah so here's the thing.... The tests will only pass if the MaxTarget is set to MaxUint256 because of this guy: [proof of work in validation](https://github.com/ubclaunchpad/cumulus/pull/112/files#diff-3a5c8ea459d421751fb70343be214598R220). 
So I dunno how to be like, if you're in test mode, MaxTarget is equal to MaxUint256, otherwise please use MaxUint232. Cuz MaxUint232 is the max target for v1.

Issue #111, opened 2017-07-26
Detect cases where we missed a broadcasted block.


bfbachmann:
We need this to be true of our implementation:
```
Block broadcasts are also tolerant of dropped
messages. If a node does not receive a block, it will request it when it receives the next block and
realizes it missed one.
```

PR #115, opened 2017-07-28
Add blockchain update mechanism, add request timeout to Peer

bfbachmann:
## Status
Ready for review

## Related Issue
#111 

## Description
* This adds a built-in timeout mechanism for peer requests. Now, when we call peer.Request and no response comes back after the default timeout period, we call the callback passed to peer.Request with a ProtocolError containing the RequestTimeout error code and message. This was users of that function don't have to implement their own timeout mechanisms as it is built into the protocol.
* Added `SyncBlockchain` function to iteratively request the next block from random peers. It is called when we boot with a target peer to connect to (download blockchain), and when we received a new block via a push message that is not valid wrt our blockchain. This way, if we are on the wrong fork of the blockchain, we miss a block, or we go offline for a while an then come back we will recognize that our blockchain is out of date and update it.
* A subtle but notable change: we now disconnect from a peer if we get 3 errors in reading it's messages over the entire course of our communication (they no longer need to be consecutive). 
* Pulled globals in App into the App struct
* Updated getLocalChain to just give us a new blockchain with a genisis block. Note that we always start with a genisis block, and if a target peer is provided we roll back our blockchain to an empty state and add the genisis block given to us by the peer.
* Added mining integration to App, as well as a `mine` flag. Mining is turned off my default. The miner restarts every time the blockchain is updated.
* Fixed a few bugs to do with decoding json messages
* Added miner command to the console for starting, stopping, and checking the status of the miner at will.
* Updated conn to use a better mock of net.Conn called BufConn 
* Fixed some bugs in Cloudbase transaction validation
* Added `BadRequest` (400) error code to message
* Broke peer testing into peer_test.go and peerstore_test.go

## WIKI Updates
* TODO

## Todos

General:
- [ ] Tests
- [ ] Wiki


app/app.go
```diff
@@ -259,3 +271,92 @@ func (a *App) HandleBlock(blk *blockchain.Block) {
 		log.Debug("added blk number %d to chain", blk.BlockNumber)
 	}
 }
+
+// SyncBlockchain updates the local copy of the blockchain by requesting missing
+// blocks from peers. Returns error if we are not connected to any peers.
+func (a *App) SyncBlockchain() error {
```
chadlagore: Nitpick, lets use `SyncBlockChain`? (since the blockchain itself is pascal case)

app/app.go
```diff
@@ -204,11 +207,21 @@ func (a *App) PushHandler(push *msg.Push) {
...
 	// TODO: Look for local chain on disk. If doesn't exist, go rummaging
 	// around on the internets for one.
-	bc, _ := blockchain.NewValidTestChainAndBlock()
-	return bc
+	bc := blockchain.BlockChain{
+		Blocks: make([]*blockchain.Block, 0),
+		Head:   blockchain.NilHash,
+	}
+
+	genisisBlock := blockchain.Genesis(user.HotWallet.Wallet.Public(),
```
chadlagore: Im not in love with the `User` API at the moment. I have some work to do there, this works for now.

app/app.go
```diff
@@ -259,3 +271,92 @@ func (a *App) HandleBlock(blk *blockchain.Block) {
...
+
+		// Get a list of peer addresses. We will use these to retreive peers from
+		// the PeerStore that we will then send block requests to
+		peerAddrs := a.PeerStore.Addrs()
+		if len(peerAddrs) == 0 {
+			return errors.New("SyncBlockchain failed: no peers to request blocks from")
+		}
+
+		// Pick a peer to send the request to (this won't always be the same one)
+		p := a.PeerStore.Get(peerAddrs[0])
```
chadlagore: Without diving deep into the peer package, its very difficult to appreciate what this line is supposed to be doing. For example, how is it that this generates a different peer each loop? Shouldn't the addresses be abstracted away at this level? Why do I care which peer I'm connecting to (or what their address is)? I just want to send block requests out to the network.

app/app.go
```diff
@@ -259,3 +271,92 @@ func (a *App) HandleBlock(blk *blockchain.Block) {
...
+		// Wait for response
+		select {
+		case newBlock := <-newBlockChan:
+			if newBlock == nil {
+				// We received a response with no error but an invalid resource
+				// Try again
+				continue
+			}
+
+			valid, validationCode := consensus.VerifyBlock(a.Chain, newBlock)
```
chadlagore: 👏 thorough!

blockchain/block.go
```diff
@@ -43,6 +43,16 @@ func (bh *BlockHeader) Marshal() []byte {
 	return buf
 }
 
+// Equal returns true if all the fields (other than ExtraData) in each of
+// the BlockHeaders match, and false otherwise.
+func (bh *BlockHeader) Equal(otherHeader *BlockHeader) bool {
```
chadlagore, line 47: Reads nicer as `Equals`

blockchain/blockchain.go
```diff
@@ -69,7 +69,7 @@ func (bc *BlockChain) ContainsTransaction(t *Transaction, start, stop uint32) (b
 	return false, 0, 0
 }
 
-// CopyLocalBlockByIndex returns a copy of a block in the local chain by index.
+// CopyBlockByIndex returns a copy of a block in the local chain by index.
```
chadlagore: 🤦‍♂️ 🙏

app/app.go
```diff
@@ -259,3 +271,92 @@ func (a *App) HandleBlock(blk *blockchain.Block) {
...
+
+// SyncBlockchain updates the local copy of the blockchain by requesting missing
+// blocks from peers. Returns error if we are not connected to any peers.
+func (a *App) SyncBlockchain() error {
+	prevHead := a.Chain.Blocks[len(a.Chain.Blocks)-1]
+	newBlockChan := make(chan *blockchain.Block)
+	errChan := make(chan *msg.ProtocolError)
+
+	// Define a handler for responses to our block requests
+	blockResponseHandler := func(blockResponse *msg.Response) {
```
chadlagore: This for example can be stubbed out for easy testing.

app/app.go
chadlagore: This is a huge function and very difficult to test. Is it possible to break it up a bit?

peer/peer.go
```diff
@@ -159,11 +161,50 @@ func (p *Peer) Dispatch() {
 
 // Request sends the given request over this peer's Connection and registers the
 // given response hadnler to be called when the response arrives at the dispatcher.
```
chadlagore: handler* - realize its old stuff

peer/peerstore_test.go
```diff
@@ -0,0 +1,253 @@
+package peer
```
chadlagore, line 1: In general, I think its a good idea to move towards the `testify` package and use assert statements. They really help with debugging, giving useful explanations of the failure (especially helpful when other people try to make changes in your code!)

chadlagore (commented):
Drive by comments. Looking good so far 🎉 

One general note about the approach we're using here to download the block chain, it seems like an unfortunately serial way to do it. I totally buy the idea of looking for the next block based on hash, but we basically do zero asynchronous torrenting from multiple peers like this (which was the goal I believe?)

blockchain/block.go
bfbachmann: I was going to make it `Equals` but that's not the Go way 🤷‍♂️.

app/app.go
bfbachmann: Yea I agree, I think there should just be a `GetRandom` function for this. But we do need to select a peer to send the request to, because broadcasting a request doesn't really make much sense.

app/app.go
```diff
@@ -204,11 +231,19 @@ func (a *App) PushHandler(push *msg.Push) {
 }
 
 // getLocalChain returns an instance of the blockchain.
-func getLocalChain() *blockchain.BlockChain {
+func getLocalChain(user *User) *blockchain.BlockChain {
```
jordanschalm: Why is user a param?

jordanschalm: It seems like the function is overloaded with two separate situations:
- we want to start a new chain with a genesis block
- we want to continue an old chain by restoring from storage

We should have two separate functions for these two situations.

app/app.go
```diff
@@ -204,11 +231,19 @@ func (a *App) PushHandler(push *msg.Push) {
...
-	return bc
+	bc := blockchain.BlockChain{
+		Blocks: make([]*blockchain.Block, 0),
+		Head:   blockchain.NilHash,
+	}
+
+	genisisBlock := blockchain.Genesis(user.Wallet.Public(),
+		consensus.CurrentTarget(), consensus.StartingBlockReward, []byte{})
+
+	bc.AppendBlock(genisisBlock)
```
jordanschalm: typo genisis -> genesis

app/app.go
```diff
@@ -236,26 +271,174 @@ func (a *App) HandleWork() {
...
-		blk := a.Pool.NextBlock(a.Chain, address, a.CurrentUser.BlockSize)
-		if miner.IsMining() {
-			miner.RestartMiner(a.Chain, blk)
+	if !validBlock {
+		// The block was invalid wrt our chain. Maybe our chain is out of date.
+		// Update it and try again.
+		chainChanged, err := a.SyncBlockChain()
+		if chainChanged && miner.IsMining() {
+			miner.StopMining()
+			go a.Mine()
```
jordanschalm: worthwhile to have a `RestartMining` method? out of scope for this PR but make an issue for it if it's worth the time

app/app.go
```diff
@@ -236,26 +271,174 @@ func (a *App) HandleWork() {
...
+			return
+		}
+	}
+}
+
+// SyncBlockChain updates the local copy of the blockchain by requesting missing
+// blocks from peers. Returns true if the blockchain changed as a result of
+// calling this function, false if it didn't and an error if we are not connected
+// to any peers.
+func (a *App) SyncBlockChain() (bool, error) {
```
jordanschalm, line 379: Can probably just have this return false if no peers and omit the error return
I'd also recommend checking for no peers before doing anything else and exiting early if we aren't connected to anything

peer/peerstore.go
```diff
@@ -156,12 +161,24 @@ func (ps *PeerStore) RemoveRandom() {
...
 	defer ps.lock.RUnlock()
 	return ps.peers[addr]
 }
 
+// GetRandom synchronously retreives a random peer from the peerstore
+// Returns nil if the PeerStore is empty
+func (ps *PeerStore) GetRandom() *Peer {
+	ps.lock.RLock()
+	defer ps.lock.RUnlock()
+	for _, p := range ps.peers {
```
jordanschalm, line 156: lol

jordanschalm (changes requested):
Very nice work.

app/app.go
bfbachmann: Sad thing is I meant to write it like that... I thought that was how it was spelled :(.

app/app.go
bfbachmann: It's possible, whoever unlikely, that all our peers disconnect from us while this function is running, in which case the blockchain might have been partially updated, but the update is incomplete. Checking if we are connected to peers before calling this would not guarantee that we have peers to make requests to throughout the duration of the function. So I think we still need both return values here.

app/app.go
```diff
@@ -33,10 +33,13 @@ const (
...
-	PeerStore   *peer.PeerStore
-	Chain       *blockchain.BlockChain
-	Pool        *pool.Pool
+	CurrentUser      *User
+	PeerStore        *peer.PeerStore
+	Chain            *blockchain.BlockChain
+	Pool             *pool.Pool
+	blockQueue       chan *blockchain.Block
+	transactionQueue chan *blockchain.Transaction
+	quitChan         chan bool
```
chadlagore, line 42: :tada:

app/app.go
```diff
@@ -80,6 +87,11 @@ func Run(cfg conf.Config) {
 	// stream in. Kick off a worker to handle requests and pushes.
 	go a.HandleWork()
 
+	if config.Mine {
```
chadlagore: This strikes me as a user-level configuration, but now I'm actually leaning towards one config file to rule them all. Perhaps we can move the `BlockSize` variable into `config` as well, then we can maybe do away with the user abstraction altogether 🤔 (Another PR)

app/app.go
```diff
@@ -149,29 +169,41 @@ func (a *App) RequestHandler(req *msg.Request) msg.Response {
...
-				res.Error = notFoundErr
-			} else {
-				res.Resource = blk
-			}
-		} else {
-			// No index parameter.
+		log.Debug("Received block request")
+
+		// Block is requested by block hash.
+		hashBytes, err := json.Marshal(req.Params["lastBlockHash"])
```
chadlagore, line 184: :tada:

app/app.go
```diff
@@ -203,12 +235,20 @@ func (a *App) PushHandler(push *msg.Push) {
 	}
 }
 
-// getLocalChain returns an instance of the blockchain.
-func getLocalChain() *blockchain.BlockChain {
-	// TODO: Look for local chain on disk. If doesn't exist, go rummaging
+// createBlockchain returns a new instance of a blockchain with only a genesis
+// block.
+func createBlockchain(user *User) *blockchain.BlockChain {
 	// around on the internets for one.
```
chadlagore: Might just blow this line away? :gun:

app/app.go
```diff
@@ -236,26 +276,169 @@ func (a *App) HandleWork() {
...
+}
+
+// Mine continuously pulls transactions form the transaction pool, uses them to
+// create blocks, and mines those blocks. When a block is mined it is added
+// to the blockchain and broadcasted into the network. Mine() returns when
+// miner.StopMining() is called.
+func (a *App) Mine() {
+	log.Info("Starting miner")
+	for {
+		// Make a new block form the transactions in the transaction pool
```
chadlagore: typo

app/app.go
```diff
@@ -236,26 +276,169 @@ func (a *App) HandleWork() {
...
+	for {
+		currentHead := a.Chain.LastBlock()
+		if currentHead == nil {
+			// Our blockchain is empty
+			currentHeadHash = blockchain.NilHash
+		} else {
+			currentHeadHash = blockchain.HashSum(currentHead)
+		}
+
+		reqParams := map[string]interface{}{
```
chadlagore: A lot of the below seems like it should be its own function.
```go
requestNextBlock(currentHeadHash)
waitForBlock()
```

app/app.go
```diff
@@ -236,26 +276,169 @@ func (a *App) HandleWork() {
...
+// calling this function, false if it didn't and an error if we are not connected
+// to any peers.
+func (a *App) SyncBlockChain() (bool, error) {
+	var currentHeadHash blockchain.Hash
+	newBlockChan := make(chan *blockchain.Block)
+	errChan := make(chan *msg.ProtocolError)
+	changed := false
+
+	// Define a handler for responses to our block requests
+	blockResponseHandler := func(blockResponse *msg.Response) {
```
chadlagore: Nitpick - I'm sure its more performant to actually define this function. Then the go complier doesn't have to re-define and allocate space for this function every time we sync the chain (which seems like it will happen a lot) - it would also shrink this function body.

app/app.go
```diff
@@ -236,26 +276,169 @@ func (a *App) HandleWork() {
...
+			return
+		}
+	}
+}
+
+// SyncBlockChain updates the local copy of the blockchain by requesting missing
+// blocks from peers. Returns true if the blockchain changed as a result of
+// calling this function, false if it didn't and an error if we are not connected
+// to any peers.
+func (a *App) SyncBlockChain() (bool, error) {
```
chadlagore, line 379: How would you test a function like this?

app/app.go
```diff
@@ -236,26 +276,169 @@ func (a *App) HandleWork() {
...
+	}
+	log.Debug("Added block number %d to chain", blk.BlockNumber)
+	return
+}
+
+// Mine continuously pulls transactions form the transaction pool, uses them to
+// create blocks, and mines those blocks. When a block is mined it is added
+// to the blockchain and broadcasted into the network. Mine() returns when
+// miner.StopMining() is called.
+func (a *App) Mine() {
```
chadlagore: Perhaps we should name this something else? Like `RunMiner`? Since we already have a fairly well defined `Mine` function.

app/app.go
```diff
@@ -203,12 +235,20 @@ func (a *App) PushHandler(push *msg.Push) {
 	}
 }
 
-// getLocalChain returns an instance of the blockchain.
-func getLocalChain() *blockchain.BlockChain {
-	// TODO: Look for local chain on disk. If doesn't exist, go rummaging
+// createBlockchain returns a new instance of a blockchain with only a genesis
+// block.
+func createBlockchain(user *User) *blockchain.BlockChain {
```
chadlagore: Is this function just for testing? Why is the current user creating the genesis block in this block chain? `getLocalChain` actually performed a function that we can improve upon, `createBlockchain` seems to deliberately create an invalid blockchain.

app/app_test.go
```diff
@@ -40,24 +42,23 @@ func TestPushHandlerNewTestTransaction(t *testing.T) {
...
 		assert.True(t, ok)
 		assert.Equal(t, tr, txn)
 	}
 }
 
-// TODO: Enable once block request by hash implemented.
-// func TestRequestHandlerNewBlockOK(t *testing.T) {
-// 	// Request a new block by hash and verify we get the right one.
-// 	a := createNewTestApp()
+func TestRequestHandlerNewBlockOK(t *testing.T) {
```
chadlagore, line 54: :tada:

app/app_test.go
```diff
@@ -71,7 +72,7 @@ func TestRequestHandlerNewBlockBadParams(t *testing.T) {
 
 	// Make sure request failed.
 	assert.False(t, ok, "resource should not contain block")
-	assert.Equal(t, msg.ResourceNotFound, int(resp.Error.Code), resp.Error.Message)
+	assert.Equal(t, msg.BadRequest, int(resp.Error.Code), resp.Error.Message)
```
chadlagore, line 78: Thanks for making these legit error messages :+1:

app/console.go
```diff
@@ -121,11 +129,41 @@ func connect(ctx *ishell.Context, a *App) {
...
-	emoji.Println(":fist: Emoji Address: " + wallet.Wallet.Public().Emoji())
+	name := shell.ReadLine()
+	wallet := blockchain.NewWallet()
+	app.CurrentUser.Wallet = wallet
+	app.CurrentUser.Name = name
+	emoji.Println(":credit_card: New wallet created!")
+	emoji.Println(":mailbox: Address: " + wallet.Public().Repr())
+}
+
+func toggleMiner(ctx *ishell.Context, app *App) {
```
chadlagore: Cool!

app/test_utils.go
```diff
@@ -7,9 +7,9 @@ import (
 	"github.com/ubclaunchpad/cumulus/pool"
 )
 
-func createNewTestBlockRequest(lastBlock interface{}) *msg.Request {
+func createNewTestBlockRequest(lastBlockHash interface{}) *msg.Request {
```
chadlagore, line 10: :tada:

blockchain/block_test.go
```diff
@@ -59,3 +58,28 @@ func TestBlockHeaderLen(t *testing.T) {
...
+		Nonce:       block1.Nonce,
+		ExtraData:   []byte("OneOrTwoOrMaybeEvenThreeOrFour"),
+	}
+
+	if !(&block1.BlockHeader).Equal(&equalBlockHeader) {
+		t.FailNow()
+	}
+
+	block2 := NewTestBlock()
+	if (&block1.BlockHeader).Equal(&block2.BlockHeader) {
```
chadlagore: We should use `assert`. Its less lines, easier to read, easier to get other people to work with your unit tests. Just far superior!

blockchain/blockchain.go
```diff
@@ -69,14 +77,31 @@ func (bc *BlockChain) ContainsTransaction(t *Transaction, start, stop uint32) (b
...
 
-// CopyLocalBlockByIndex returns a copy of a block in the local chain by index.
-func (bc *BlockChain) CopyBlockByIndex(i uint32) (*Block, error) {
-	if i >= 0 && i < uint32(len(bc.Blocks)) {
-		blk := bc.Blocks[i]
-		b := *blk
-		b.Transactions = make([]*Transaction, len(blk.Transactions))
-		copy(b.Transactions, blk.Transactions)
-		return &b, nil
+// GetBlockByLastBlockHash returns a copy of the block in the local chain that
```
chadlagore, line 80: not a copy

blockchain/blockchain.go
```diff
@@ -69,14 +77,31 @@ func (bc *BlockChain) ContainsTransaction(t *Transaction, start, stop uint32) (b
...
+		if block.LastBlock == hash {
+			return block, nil
+		}
+	}
+	return nil, errors.New("No such block")
+}
+
+// RollBack removes the last block from the blockchain. Returns the block that
+// was removed from the end of the chain, or nil if the blockchain is empty.
+func (bc *BlockChain) RollBack() *Block {
```
chadlagore, line 95: Interesting

blockchain/blockchain.go
```diff
@@ -69,14 +77,31 @@ func (bc *BlockChain) ContainsTransaction(t *Transaction, start, stop uint32) (b
...
+	if len(bc.Blocks) == 0 {
+		return nil
+	}
+	prevHead := bc.LastBlock()
+	bc.Blocks = bc.Blocks[:len(bc.Blocks)-1]
```
chadlagore, line 100: Fancy golang pop:

```
prevHead, bc.Blocks = bc.Blocks[len(bc.Blocks)-1], bc.Blocks[:len(bc.Blocks)-1]
```

https://github.com/golang/go/wiki/SliceTricks

chadlagore: Yours is probably more readable 😆

chadlagore (commented):
Wow, lots going on here! Couple notes:

1. `assert` everywhere, literally saves you 50 lines in this pr and saves everyone else some serious head scratching. There is no reason not to use it at this point.

2. Long functions are hard to test, especially in the app package which already suffers from low coverage (it makes it really hard for others to make changes to your code in the future, the `Run` function is totally untested, and now the `SyncBlockChain` logic appears to be going the same way).

3. The scope is so far blown out, I don’t even see where this PR addresses ticket 111? Three separate PRs here would be better than one if you want legit feedback - its actually difficult to sign off on all of this because it takes so long to grok it all! Plus you expose yourself to too many TODO’s from onlookers. Takes forever to merge.

Theres lots about this PR that is sweet, but I :pray: to see two things before we merge: 

- [ ] Reduce the complexity of `SyncBlockChain` so that it doesn't go forever untested. Test it piece by piece.
- [ ] Use `assert`. It would take 10mins to swap these `t.FailNow`'s out.

**EDIT**: Im noticing that this is probably more geared towards issue 118? In that case the scope makes more sense! I thought this was literally random coding lol.

app/app.go
bfbachmann: The Cobra library we're using already supports config files by default.

app/app.go
bfbachmann: Not invalid, just a genesis block, so it's valid. I got rid of `getLocalChain` because it was trying to do 2 things (create new blockchain or load one from disk). I'm breaking that into two functions, one is `createBlockchain` and the other will come when we support loading from disk.

app/app.go
bfbachmann: Hard to do, but it would have to be an integration tests where the App instance has a peerstore with connected peers who have copies of the blockchain. I don't see a way to make this function less complex without making it useless.

chadlagore: Break it into 5 functions?

app/app.go
```diff
@@ -343,15 +338,22 @@ func (a *App) Mine() {
 	}
 }
 
+// RestartMiner restarts the miner only if it is running when this is called.
+func (a *App) RestartMiner() {
```
chadlagore: 🎉

chadlagore (approved):
Looks great!

Issue #101, opened 2017-07-15
Add transaction command


jordanschalm:
ping @chadlagore what is the status on this?

chadlagore:
https://github.com/ubclaunchpad/cumulus/pull/120

PR #120, opened 2017-08-10
Add Transaction Command

chadlagore:
**Status**: _Open for visibility_

## Related Issue

[101](https://github.com/ubclaunchpad/cumulus/issues/101)

## Description

* Implementing the add transaction functionality.
* Has some wide ranging effects: `Transactions`, `Wallets`, `console` etc.
* Adjusted the transaction recipient to only use checksums instead of `Address`.
* We may need to do the same with the transaction sender ☝️ 
* Broadcast new transactions to other nodes.

```bash
What would you like to create?
   Wallet
 ❯ Transaction
💳  Enter recipient wallet address
>>> ef653f0f6f292b7a12392686ea03058045ad51c1
💵  Enter amount to send:
>>> 15
📬  Its in the mail!
>>> _
```

## WIKI Updates
* [Using the Console](https://github.com/ubclaunchpad/cumulus/wiki/Using-the-Console)

## Todos
- [x] Initialize balance.
- [x] Test new `Wallet` methods.
- [x] Update the `Wallet` when new blocks come in to kill pending transactions.
- [x] Broadcast transactions to other nodes
- [x] Update own pool.

General:
- [x] Tests
- [x] Wiki

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -36,7 +36,7 @@ func (a *App) Pay(to string, amount uint64) error {
 	wallet := a.CurrentUser.Wallet
 	pool := a.Pool
 
-	// 1. A legitimate transaction must be built.
+	// A legitimate transaction must be built.
 	tbody := blockchain.TxBody{
 		Sender: wallet.Public(),
 		// TODO: Collect inputs.
@@ -49,22 +49,24 @@ func (a *App) Pay(to string, amount uint64) error {
 		},
 	}
 
+	// The transaction must be signed.
 	if txn, err := tbody.Sign(*a.CurrentUser.Wallet, crand.Reader); err == nil {
-		// 2. The transaction must be signed.
-		a.PeerStore.Broadcast(msg.Push{
-			ResourceType: msg.ResourceTransaction,
-			Resource:     txn,
-		})
 
-		// 4. The transaction must be broadcasted to the peers.
+		// The transaction must be broadcasted to the peers.
 		if err := wallet.SetPending(txn); err != nil {
 			return err
 		}
 
-		// 3. The transaction must be added to the pool.
+		// The transaction must be added to the pool.
 		if !pool.Push(txn, a.Chain) {
 			return errors.New("transaction broadcasted, but could not be added to the pool")
 		}
+
+		// The transaction must be broadcasted to the network.
+		a.PeerStore.Broadcast(msg.Push{
+			ResourceType: msg.ResourceTransaction,
+			Resource:     txn,
+		})
 
 	} else {
 		return err
--- a/app/user.go
+++ b/app/user.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"github.com/ubclaunchpad/cumulus/blockchain"
+	"github.com/ubclaunchpad/cumulus/msg"
 
 	crand "crypto/rand"
 )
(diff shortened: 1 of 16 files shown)


app/user.go
```diff
@@ -1,33 +1,61 @@
...
-// getCurrentUser gets the current user.
+// getCurrentUser gets the current user function only used for app initalization.
 func getCurrentUser() *User {
 	// TODO: Check for local user information on disk,
 	// If doesnt exist, create new user.
 	return NewUser()
 }
+
+// Pay pays an amount of coin to an address `to`.
+func (a *App) Pay(to string, amount uint64) error {
```
jordanschalm, line 34: I think we need to make some changes to this function.

We should:
1. Create a signed transaction
2. Broadcast it to all connected nodes

Independently of this function, the balance of the current user's wallet should be computed each time a new block is added to our chain.

blockchain/wallet.go
```diff
@@ -93,27 +97,32 @@ func (a Address) Key() *ecdsa.PublicKey {
 	}
 }
 
-// Wallet represents a wallet that we have the ability to sign for.
-type Wallet interface {
+// Account represents a wallet that we have the ability to sign for.
+type Account interface {
 	Public() Address
 	Sign(digest Hash, random io.Reader) (Signature, error)
+	Debit(amount uint64) error
```
jordanschalm: Similar to above, I don't think we should modify a user account directly, it should be a function of the blockchain and should be recomputed each time we add a block to the chain.

jordanschalm (commented):
Really like the create transaction flow. We need to consider how we compute wallet balances, but we can leave that to a different PR. It's quite a large task and out of scope of this issue.

app/user.go
chadlagore: Good point, so the wallet balance is only updated when our transaction has landed in a block - makes total sense.

jordanschalm (approved):
Looking good

app/user.go
```diff
@@ -1,33 +1,76 @@
...
+				Recipient: to,
+				Amount:    amount,
+			},
+		},
+	}
+
+	// The transaction must be signed.
+	if txn, err := tbody.Sign(*a.CurrentUser.Wallet, crand.Reader); err == nil {
+
+		// The transaction must be broadcasted to the peers.
```
bfbachmann, line 55: The comment here doesn't fit what's happening

chadlagore: Fixed up in inputs PR 😅

blockchain/wallet.go
```diff
@@ -147,9 +156,82 @@ func (s *Signature) Marshal() []byte {
...
+	}
+	return false, -1
+}
+
+// GetEffectiveBalance returns the wallet balance less the sum of the pending
+// transactions in the wallet.
+func (w *Wallet) GetEffectiveBalance() uint64 {
+	r := w.Balance
+	for _, t := range w.PendingTxns {
+		r -= t.GetTotalOutput()
```
bfbachmann, line 223: Is this mathematically correct? Won't part of the output to each transaction go back to the sender, meaning the total output is not actually how much you're spending?

consensus/consensus.go
```diff
@@ -82,7 +82,7 @@ func VerifyCloudBase(bc *blockchain.BlockChain,
 	reward := StartingBlockReward / uint64(math.Pow(float64(2), timesHalved))
 
 	// Check that the output is properly set.
-	if len(t.Outputs) != 1 || t.Outputs[0].Recipient == blockchain.NilAddr {
+	if len(t.Outputs) != 1 || t.Outputs[0].Recipient == blockchain.NilAddr.Repr() {
```
bfbachmann: This could throw an index out of bounds error if there are no outputs on the transaction.

Issue #126, opened 2017-08-13
Allow multiple inputs in transactions


jordanschalm:
Modify transactions to allow multiple inputs. Update consensus rules for this change

PR #127, opened 2017-08-13
Pluralized Transaction Inputs

chadlagore:
**Status**: _Ready for review_

## Related Issue
[126](https://github.com/ubclaunchpad/cumulus/issues/126)

## Description
* Transactions should reference multiple inputs.
* This allows users to coalesce micro-payments into larger ones.

## TODOs
- [x] Pluralize transactions
- [x] Get tests passing again.
- [x] Rebase brunos branch.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -351,7 +351,7 @@ func (a *App) RunMiner() {
 		miningResult := miner.Mine(a.Chain, blockToMine)
 
 		if miningResult.Complete {
-			// log.Debug("Successfully mined a block!")
+			log.Info("Successfully mined a block!")
 			push := msg.Push{
 				ResourceType: msg.ResourceBlock,
 				Resource:     blockToMine,
--- a/app/app.go
+++ b/app/app.go
@@ -284,8 +284,8 @@ func (a *App) HandleWork() {
 
 // HandleTransaction handles new instance of TransactionWork.
 func (a *App) HandleTransaction(txn *blockchain.Transaction) {
-	validTransaction := a.Pool.Push(txn, a.Chain)
-	if validTransaction {
+	code := a.Pool.Push(txn, a.Chain)
+	if code == consensus.ValidTransaction {
 		log.Debug("Added transaction to pool from address: " + txn.Sender.Repr())
 	} else {
 		log.Debug("Bad transaction rejected from sender: " + txn.Sender.Repr())
@@ -351,7 +351,7 @@ func (a *App) RunMiner() {
 		miningResult := miner.Mine(a.Chain, blockToMine)
 
 		if miningResult.Complete {
-			log.Info("Successfully mined a block!")
+			// log.Debug("Successfully mined a block!")
 			push := msg.Push{
 				ResourceType: msg.ResourceBlock,
 				Resource:     blockToMine,
@@ -457,7 +457,7 @@ func (a *App) makeBlockRequest(currentHead *blockchain.Block,
 
 // handleBlockResponse receives a block or nil from the newBlockChan and attempts
 // to validate it and add it to the blockchain, or it handles a protocol error
-// from the errChan. Returns whether the blockchain was modified and wether we
+// from the errChan. Returns whether the blockchain was modified and whether we
 // received an UpToDate response.
 func (a *App) handleBlockResponse(newBlockChan chan *blockchain.Block,
 	errChan chan *msg.ProtocolError) (changed bool, upToDate bool) {
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -47,7 +47,7 @@ func TestPushHandlerNewTestTransaction(t *testing.T) {
 	select {
 	case tr, ok := <-a.transactionQueue:
 		assert.True(t, ok)
-		assert.Equal(t, tr, txn)
+		assert.ObjectsAreEqual(tr, txn)
 	}
 }
 
@@ -138,7 +138,7 @@ func TestHandleValidBlock(t *testing.T) {
 	bc, blk := blockchain.NewValidTestChainAndBlock()
 	a.Chain = bc
 	a.HandleBlock(blk)
-	assert.Equal(t, blk, a.Chain.Blocks[2])
+	assert.ObjectsAreEqual(blk, a.Chain.Blocks[2])
 
 	// TODO: Assert miner restarted.
 	// TODO: Assert pool appropriately emptied.
@@ -226,14 +226,24 @@ func TestHandleBlockResponse(t *testing.T) {
 	a := createNewTestApp()
 	newBlockChan := make(chan *blockchain.Block, 1)
 	errChan := make(chan *msg.ProtocolError, 1)
-	newBlockChan <- a.Chain.RollBack()
+
+	// Roll the last block off of the chain, add to channel.
+	lastBlock := a.Chain.RollBack()
+	assert.NotNil(t, lastBlock)
+	newBlockChan <- lastBlock
+
+	// Handle the block.
 	changed, upToDate := a.handleBlockResponse(newBlockChan, errChan)
 	assert.True(t, changed)
 	assert.False(t, upToDate)
+
+	// Add an on-chain block to the handler (this shouldn't change the chain).
 	newBlockChan <- a.Chain.Blocks[1]
 	changed, upToDate = a.handleBlockResponse(newBlockChan, errChan)
 	assert.False(t, changed)
 	assert.False(t, upToDate)
+
+	// More stuff happens.
 	errChan <- msg.NewProtocolError(msg.UpToDate, "")
 	changed, upToDate = a.handleBlockResponse(newBlockChan, errChan)
 	assert.False(t, changed)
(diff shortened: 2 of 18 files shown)


blockchain/blockchain.go
```diff
@@ -80,3 +90,25 @@ func (bc *BlockChain) CopyBlockByIndex(i uint32) (*Block, error) {
 	}
 	return nil, errors.New("block request out of bounds")
 }
+
+// InputsSpentElsewhere returns true if inputs perported to be only spent
```
jordanschalm: purported

blockchain/transaction.go
```diff
@@ -112,3 +116,53 @@ func (t *Transaction) GetTotalOutput() uint64 {
...
+		result += in.GetTotalOutputFor(t.Sender.Repr())
+	}
+	return result
+}
+
+// GetBlockRange returns the start and end block indexes for the inputs
+// to a transaction.
+func (bc *BlockChain) GetBlockRange(t *Transaction) (uint32, uint32) {
+	min := uint32(math.MaxUint32)
+	max := uint32(math.MaxUint32) // Why I have to cast this? No idea.
```
jordanschalm: should be 0?

consensus/consensus.go
```diff
@@ -223,9 +222,8 @@ func VerifyBlock(bc *blockchain.BlockChain,
 	// Check for multiple transactions referencing same input transaction.
 	for i, trA := range b.Transactions {
```
jordanschalm: What about:
```
create input map M
for t in transactions:
  for inp in t.inputs:
    attempt to add inp to M
    if already in M: fail
```

pool/pool.go
```diff
@@ -53,7 +53,8 @@ func (p *Pool) GetN(N int) *blockchain.Transaction {
 
 // GetIndex returns the index of the transaction in the ordering.
 func (p *Pool) GetIndex(t *blockchain.Transaction) int {
```
jordanschalm, line 53: We lose the property that attempting to add a transaction with an input that already exists will fail. Now we can do [IN1, IN2]->OUT1 and [IN1, IN3]->OUT2 and addition will succeed. What are we doing to mediate this and ensure we don't add to the pool if an input already exists? A many-to-one mapping for inputs -> transactions would be nice. Like IN1->OUT1, IN2->OUT2.

jordanschalm (commented):
Looking good so far 👍

blockchain/transaction.go
chadlagore: Meant to use `MinUint32` - nice catch!

pool/pool.go
chadlagore: Its actually ok that multiple transactions in the pool exist with the same input. A single transaction may have awarded 2 coins to alice and 1 to bob, that transaction will serve as an input in at most 2 transactions in the future.

I'll admit that this implementation isn't quite perfect though 😬.

chadlagore: The same situation goes for testing double spends within blocks. Its ok that an input is referenced more than once, but not if it is referenced more than once _by the same sender_. Thanks to your map idea, we now do that check in linear time though :tada:

app/app.go
```diff
@@ -351,7 +351,7 @@ func (a *App) RunMiner() {
 		miningResult := miner.Mine(a.Chain, blockToMine)
 
 		if miningResult.Complete {
-			log.Info("Successfully mined a block!")
+			// log.Debug("Successfully mined a block!")
```
chadlagore: Quite noisy in tests!

bfbachmann: We can just set logging levels in individual tests or in a `TestMain`.

app/app_test.go
```diff
@@ -47,7 +47,7 @@ func TestPushHandlerNewTestTransaction(t *testing.T) {
 	select {
 	case tr, ok := <-a.transactionQueue:
 		assert.True(t, ok)
-		assert.Equal(t, tr, txn)
+		assert.ObjectsAreEqual(tr, txn)
```
bfbachmann: Have to use `Equal` :(

app/app_test.go
```diff
@@ -242,19 +252,30 @@ func TestHandleBlockResponse(t *testing.T) {
...
-	time.Sleep(time.Second)
-	assert.Equal(t, len(a.Chain.Blocks), 1)
+
+	// Roll the last block off of the chain, add to channel
+	// (expect it not to be added back).
+	lastBlock = a.Chain.RollBack()
+	assert.Equal(t, len(a.Chain.Blocks), 2)
+	a.blockQueue <- lastBlock
+	time.Sleep(time.Millisecond * 50)
+	assert.Equal(t, len(a.Chain.Blocks), 2)
```
bfbachmann, line 289: Nice tests!

blockchain/block.go
```diff
@@ -69,7 +69,7 @@ func (b *Block) Len() int {
 }
 
 // Marshal converts a Block to a byte slice.
-func (b *Block) Marshal() []byte {
+func (b Block) Marshal() []byte {
```
bfbachmann, line 72: I actually think it's better to use the pointer receiver here because that way we don't have to copy the block when we call this.

blockchain/blockchain.go
```diff
@@ -53,17 +53,32 @@ func (bc *BlockChain) LastBlock() *Block {
...
 		return nil
 	}
-	return b.Transactions[t.Input.Index]
+	return b.Transactions[t.Index]
+}
+
+// GetAllInputs returns all the transactions referenced by a transaction
+// as inputs. Returns an error if any of the transactios requested could
+// not be found.
+func (bc *BlockChain) GetAllInputs(t *Transaction) ([]*Transaction, error) {
```
bfbachmann, line 103: Lovely

blockchain/test_utils.go
```diff
@@ -123,82 +127,217 @@ func NewTestOutputBlock(t []*Transaction, input *Block) *Block {
...
+	cbB, _ := NewValidCloudBaseTestTransaction()
+	cbC, _ := NewValidCloudBaseTestTransaction()
 
-	trB, _ := NewTestTransactionValue(sender, recipient, 4, 1)
-	trB.Input.Hash = HashSum(trA)
+	// Transaction A is at index 1 in block 0 (sender awarded 4 coins).
+	tA, _ := TxBody{
+		Sender: sender.Public(),
+		Inputs: []TxHashPointer{},
+		Outputs: []TxOutput{
```
bfbachmann, line 169: It might be a good idea to enforce the condition that at least one of the outputs from transaction goes to someone other than the sender.

blockchain/transaction.go
```diff
@@ -112,3 +116,93 @@ func (t *Transaction) GetTotalOutput() uint64 {
...
+		}
+		if in.BlockNumber > max {
+			max = in.BlockNumber
+		}
+	}
+	return min, max
+}
+
+// InputsIntersect returns true if the inputs of t intersect with those of other.
+func (t *Transaction) InputsIntersect(other *Transaction) bool {
```
bfbachmann, line 166: ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

consensus/consensus.go
```diff
@@ -61,10 +64,10 @@ func VerifyCloudBase(bc *blockchain.BlockChain,
 		return false, BadCloudBaseSender
 	}
 
-	// Check that the input is 0.
-	if t.TxBody.Input.BlockNumber != 0 ||
-		t.TxBody.Input.Hash != blockchain.NilHash ||
-		t.Input.Index != 0 {
+	// Check that the input is 0 (only one input to CB).
+	input := t.TxBody.Inputs[0]
```
bfbachmann: Might want to guard against an out-of-bounds error here.

consensus/consensus.go
```diff
@@ -170,14 +173,14 @@ func VerifyBlock(bc *blockchain.BlockChain,
 	}
 
 	// Check that block number is valid.
-	ix := b.BlockNumber - 1
-	if int(ix) > len(bc.Blocks)-1 || ix < 0 {
+	lastBlockIdx := b.BlockNumber - 1
+	if int(lastBlockIdx) != len(bc.Blocks)-1 {
```
bfbachmann, line 179: I would change the comment on this function to say something about the fact that we are making sure that this block can be added to the _end_ of the chain.

miner/miner.go
```diff
@@ -120,8 +120,8 @@ func CloudBase(
 	// Increment the input index of every transaction that has an input in the
 	// new block
 	for _, tx := range b.Transactions[1:] {
```
bfbachmann: I think this for loop should be removed because a transaction can't reference an input in the same block.

bfbachmann (approved):
Extremely high quality 🏅

blockchain/test_utils.go
chadlagore: I'll sleep on this 😴

PR #134, opened 2017-09-02
125 save user info

### LINKED ISSUE #125: Save user info to disk
On shutdown, save user data to disk.
On startup, try reading user data from disk.

### LINKED ISSUE #131: Save blockchain to disk

bfbachmann:
## Related Issue
Closes #125
Closes #131

## Description
* Added `Save` and `Load` functions to user and blockchain to save and load their contents from disk on startup.
* Added a `user` command to the console to view and edit user info.

## WIKI Updates
* TODO

## Todos

General:
- [x] Tests
- [x] Documentation
- [ ] Wiki

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -29,7 +29,7 @@ func NewUser() *User {
 	}
 }
 
-// Public returns the public key (address) of the given user
+// Public returns the public key of the given user
 func (u *User) Public() blockchain.Address {
 	return u.Wallet.Public()
 }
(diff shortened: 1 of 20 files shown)


app/user.go
```diff
@@ -1,36 +1,78 @@
...
 func NewUser() *User {
 	return &User{
-		Wallet:    *blockchain.NewWallet(),
+		Wallet:    blockchain.NewWallet(),
 		BlockSize: blockchain.DefaultBlockSize,
+		Name:      "Default User",
+	}
+}
+
+// Public returns the public key (address) of the given user
```
jordanschalm: Public key and address are not the same. Address is hash of key, is that right @chadlagore ?

app/user.go
```diff
@@ -1,36 +1,78 @@
...
-// getCurrentUser gets the current user function only used for app initalization.
-func getCurrentUser() *User {
-	// TODO: Check for local user information on disk,
-	// If doesnt exist, create new user.
-	return NewUser()
+// Load attempts to read user info from the file with the given name in the
+// current working directory in JSON format. On success this returns
+// a pointer to a new user constructed from the information in the file.
+// If an error occurrs it is returned.
+func Load(fileName string) (*User, error) {
```
jordanschalm: I'd call this `LoadUser` since it's in the `app` namespace. `app.Load` seems like it should load the app

jordanschalm (approved):

app/user.go
bfbachmann: Good call, I'll remove that "(address)" comment, it's not really necessary.

app/user.go
bfbachmann: Will do.

PR #135, opened 2017-09-17
101 transaction command

bfbachmann:
_Please review and merge #125 before reviewing this PR_

## Related Issue
#101 

## Description
I added quite a bit of functionality in this PR
* The key thing I added was the transaction command. This required me to make changes to `wallet`, `user`, `blockchain`, `app`, and `console`.
* App:
    - We now propagate new transactions through the network if we havn't already seen them
    - Fixed bug where we don't update the user's wallet after synchronizing the blockchain
* Console:
    - Minor print formatting improvements
    - Add transaction command
* User:
    - Fixed bugs in collectInputsForTransaction()
    - Update Pay() function and implemented collecting inputs for new transactions
* Block:
    - Fixed bug in getTotalInputFrom()
    - Added supporting functions for getting transactions/ transaction amounts to and from certain senders in a block
* Blockchain:
    - Moved constants from current.go to blockchain.go
* Wallet:
    - Remove wallet balance manipulation from DropAllPending(). That is now handled separately
    - Fix bug in GetEffectiveBalance() calculations
    - Add Refresh() to update the wallet balance from a partially/entirely new blockchain
* Pool:
    - Fixed SIGSEGV risk in Get()

## WIKI Updates
None. I should probably do that at some point.

## Todos

General:
- [X] Tests
- [X] Documentation
- [ ] Wiki

### DIFF
--- a/app/console.go
+++ b/app/console.go
@@ -1,12 +1,11 @@
 package app
 
 import (
-	"encoding/json"
+	"fmt"
 	"strconv"
 
 	"github.com/abiosoft/ishell"
 	"github.com/ubclaunchpad/cumulus/blockchain"
-	"github.com/ubclaunchpad/cumulus/consensus"
 	"github.com/ubclaunchpad/cumulus/miner"
 	"github.com/ubclaunchpad/cumulus/peer"
 	"gopkg.in/kyokomi/emoji.v1"
@@ -98,14 +97,14 @@ func send(ctx *ishell.Context, app *App) {
 		ctx.Println("Amount must be a positive decimal value")
 		return
 	}
-	amount *= float64(consensus.CoinValue)
+	amount *= float64(blockchain.CoinValue)
 	addr := ctx.Args[1]
 
 	// Try to make a payment.
+	ctx.Println("Sending amount", coinValue(uint64(amount)), "to", addr)
 	err = app.Pay(addr, uint64(amount))
 	if err != nil {
-		emoji.Print(":disappointed: ")
-		ctx.Println(err.Error())
+		emoji.Println(":disappointed: ", err)
 	} else {
 		emoji.Println(":mailbox_with_mail: Its in the mail!")
 	}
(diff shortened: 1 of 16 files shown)


.gitignore
```diff
@@ -35,3 +35,7 @@ vendor
 
 # Log files
 logfile
+
+# Saved app data
+user.json
```
jordanschalm: This should be (a) configurable and (b) default to `Library` or something. Can be done later but add an issue

app/console.go
```diff
@@ -61,33 +65,80 @@ func RunConsole(a *App) *ishell.Shell {
...
+
+	amount, err := strconv.ParseFloat(ctx.Args[0], 64)
+	if err != nil {
+		ctx.Println(err)
+		return
+	} else if amount <= 0 {
+		ctx.Println("Amount must be a positive decimal value")
+		return
+	}
+	amount *= float64(blockchain.CoinValue)
```
jordanschalm: I think this conversion is more complicated and probably requires a helper function to deal with edge cases

app/console.go
```diff
@@ -170,27 +228,44 @@ func createWallet(ctx *ishell.Context, app *App) {
...
+		ctx.Println("Address:", app.CurrentUser.Public().Repr())
+		emoji.Println("Emoji Address:", app.CurrentUser.Public().Emoji())
+	} else if len(ctx.Args) == 2 {
+		if ctx.Args[0] == "name" {
+			app.CurrentUser.Name = ctx.Args[1]
+			if err := app.CurrentUser.Save(userFileName); err != nil {
+				ctx.Print(err)
+			}
+			return
+		} else if ctx.Args[0] == "blocksize" {
```
jordanschalm: Out of scope for this PR but blocksize absolutely should not be editable by the user this way. Consensus needs to be consistent across all nodes

app/user.go
```diff
@@ -52,29 +101,101 @@ func (a *App) Pay(to string, amount uint64) error {
...
+		txns := block.GetTransactionsTo(sender)
+
+		// Add all the transactions to sender to our list of potential input
+		// transactions until the total is greater that or equal to the amount
+		// for the transaction we want to send.
+		for i, txn := range txns {
+			if i != 0 {
+				// This transaction is not a CloudBase transaction. Add it's
+				// inputs to a map of spent transactions so we don't spend them.
+				for _, hashPtr := range txn.Inputs {
```
jordanschalm: Maybe only add spent TXNs to the map if they _could_ be used by us. The proportion of transactions to any given user will be small

blockchain/test_utils.go
```diff
@@ -363,7 +318,7 @@ func NewValidCloudBaseTestTransaction() (*Transaction, Address) {
 		Index:       0,
 	}
 	cbReward := TxOutput{
-		Amount:    25,
+		Amount:    StartingBlockReward,
```
jordanschalm: 👍

jordanschalm (approved):
Also, pretty much all of your `it's` should be `its` 😛

Issue #136, opened 2017-09-28
Enhance Private Key Storage Security


david-julien:
- generate keys from a seed phrase, so you can use the phrase to recover the key
- require a password from user to encrypt key
- store key encrypted WHEREVER POSSIBLE
- ask user for password whenever key needs to be unencrypted to sign something

PR #137, opened 2017-12-04
133 secure private key storage

david-julien:
## Related Issue
[https://github.com/ubclaunchpad/cumulus/issues/136]()

## Description
* The goal of this PR is to encrypt the user's private key at rest using a password.
* Users will be able to enable and disable a cryptowallet
* If the cryptowallet is enabled, users will be required to submit a password before sending cumuli

## WIKI Updates
* [link]()

## Todos

General:
- [ ] Tests
- [ ] Documentation
- [x] Wiki

Other:
* Password complexity verification
* Re-enter password before encrypting

### DIFF
--- a/app/console.go
+++ b/app/console.go
@@ -110,7 +110,7 @@ func encrypt(ctx *ishell.Context, app *App, password string) error {
 func decrypt(ctx *ishell.Context, app *App, password string) error {
 	err := app.CurrentUser.DecryptPrivateKey(password)
 	if err != nil {
-		log.Printf("Unable to decrypt private key: %s", err.Error())
+		//log.Printf("Unable to decrypt private key: %s", err.Error())
 		return err
 	}
 
@@ -132,34 +132,46 @@ func cryptoWallet(ctx *ishell.Context, app *App) {
 	switch ctx.Args[0] {
 	case "enable":
 		if app.CurrentUser.CryptoWallet {
-			ctx.Print("CryptoWallet is already enabled.")
+			ctx.Print("CryptoWallet is already enabled")
 		} else {
 			ctx.Print("Please enter password: ")
 			password := ctx.ReadPassword()
 			err := encrypt(ctx, app, password)
 			if err != nil {
-				ctx.Println("Unable to decrypt private key")
+				ctx.Print("Unable to decrypt private key")
+			} else {
+				ctx.Print("Successfully enabled cryptowallet")
 			}
 		}
 	case "disable":
 		if !app.CurrentUser.CryptoWallet {
-			ctx.Print("CryptoWallet is already disabled.")
+			ctx.Print("CryptoWallet is already disabled")
 		} else {
-			ctx.Println("Please enter password: ")
+			ctx.Print("Please enter password: ")
 			password := ctx.ReadPassword()
 			err := decrypt(ctx, app, password)
 			if InvalidPassword(err) {
-				ctx.Println("Inavalid password, please try again: ")
+				ctx.Print("Inavalid password, please try again: ")
 				password := ctx.ReadPassword()
 				err := decrypt(ctx, app, password)
 				if err != nil {
 					ctx.Println("Unable to decrypt private key")
 				}
+			} else {
+				ctx.Print("Successfully disabled cryptowallet")
 			}
 
 		}
+	case "status":
+		var s string
+		if app.CurrentUser.CryptoWallet {
+			s = "enabled"
+		} else {
+			s = "disabled"
+		}
+		ctx.Printf("cryptowallet status: %s", s)
 	default:
-		ctx.Println("")
+		//Fall through
 	}
 }
 
--- a/app/console.go
+++ b/app/console.go
@@ -2,7 +2,6 @@ package app
 
 import (
 	"fmt"
-	"log"
 	"strconv"
 	"strings"
 
@@ -92,10 +91,9 @@ func RunConsole(a *App) *ishell.Shell {
 	return shell
 }
 
-func encrypt(ctx *ishell.Context, app *App, password string) error {
+func encryptUser(ctx *ishell.Context, app *App, password string) error {
 	err := app.CurrentUser.EncryptPrivateKey(password)
 	if err != nil {
-		log.Printf("Unable to encrypt private key: %s", err.Error())
 		return err
 	}
 
@@ -108,10 +106,9 @@ func encrypt(ctx *ishell.Context, app *App, password string) error {
 	return nil
 }
 
-func decrypt(ctx *ishell.Context, app *App, password string) error {
+func decryptUser(ctx *ishell.Context, app *App, password string) error {
 	err := app.CurrentUser.DecryptPrivateKey(password)
 	if err != nil {
-		//log.Printf("Unable to decrypt private key: %s", err.Error())
 		return err
 	}
 
@@ -126,7 +123,10 @@ func decrypt(ctx *ishell.Context, app *App, password string) error {
 
 func cryptoWallet(ctx *ishell.Context, app *App) {
 	if len(ctx.Args) < 1 {
-		ctx.Println("Usage: cryptowallet [enable/disable]")
+		ctx.Println("\nUsage: cryptowallet [command]")
+		ctx.Println("\nCOMMANDS:")
+		ctx.Println("\t enable \t Enable cryptowallet")
+		ctx.Println("\t disable \t Disable cryptowallet")
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


app/console.go
```diff
@@ -72,6 +74,13 @@ func RunConsole(a *App) *ishell.Shell {
 			editUser(ctx, a)
 		},
 	})
+	shell.AddCmd(&ishell.Cmd{
+		Name: "cryptowallet",
```
chadlagore, line 76: Fingerbreakingly painful command 😂, how about encrypt?

bfbachmann: Havnt has a chance to look at the whole PR but I prefer cryptowallet to encrypt. It might be slightly longer but it’s more accurate. Plus the CLI has autocomplete.

david-julien: Yeah encrypt is a little vague to the user. What am I encrypting? But I agree it would be better shorter if anyone can think of anything

chadlagore: wallet encrypt? I don’t think cryptowallet is very clear. Maybe I’m splitting hairs.

app/console.go
```diff
@@ -83,6 +92,92 @@ func RunConsole(a *App) *ishell.Shell {
...
+		ctx.Print(err)
+		return err
+	}
+
+	return nil
+}
+
+func cryptoWallet(ctx *ishell.Context, app *App) {
+	if len(ctx.Args) < 1 {
+		ctx.Println("Usage: cryptowallet [enable/disable]")
```
bfbachmann: Minor: I think we should use the usage documentation style I use in `toggleMiner()` just for consistency. Should look something like
```go
ctx.Println("\nUsage: cryptowallet [command]")
ctx.Println("\nCOMMANDS:")
ctx.Println("\t enable \t Enable cryptowallet")
ctx.Println("\t disable \t Disable cryptowallet")
```

app/console.go
```diff
@@ -2,7 +2,9 @@ package app
 
 import (
 	"fmt"
+	"log"
```
bfbachmann: Use package `github.com/Sirupsen/logrus` for logging.

app/console.go
```diff
@@ -83,6 +92,92 @@ func RunConsole(a *App) *ishell.Shell {
 	return shell
 }
 
+func encrypt(ctx *ishell.Context, app *App, password string) error {
+	err := app.CurrentUser.EncryptPrivateKey(password)
+	if err != nil {
+		log.Printf("Unable to encrypt private key: %s", err.Error())
```
bfbachmann: With Logrus this should be `log.Errorf("Unable to encrypt private key: %s", err.Error())`, but don't log in this function. You are returning the error to your caller so it's better to let them handle the logging for you. If you log low down the call stack when you're propagating the error upward anyway you risk logging a bunch of duplicate errors (because your callers might log the errors you return to them as well). Just return the error in this function and in `decrypt`, and do the logging and console output in `cryptoWallet()`.

app/console.go
```diff
@@ -83,6 +92,92 @@ func RunConsole(a *App) *ishell.Shell {
...
+		return err
+	}
+
+	return nil
+}
+
+func decrypt(ctx *ishell.Context, app *App, password string) error {
+	err := app.CurrentUser.DecryptPrivateKey(password)
+	if err != nil {
+		//log.Printf("Unable to decrypt private key: %s", err.Error())
```
bfbachmann: Sorry, no commented-out code allowed.

app/console.go
```diff
@@ -83,6 +92,92 @@ func RunConsole(a *App) *ishell.Shell {
...
+	case "status":
+		var s string
+		if app.CurrentUser.CryptoWallet {
+			s = "enabled"
+		} else {
+			s = "disabled"
+		}
+		ctx.Printf("cryptowallet status: %s", s)
+	default:
+		//Fall through
```
bfbachmann: Might want to print usage here too in case someone tries something like `cryptoWallet dkljhgd nslufgdli`.

app/console.go
```diff
@@ -100,9 +195,36 @@ func send(ctx *ishell.Context, app *App) {
...
+			return
+		}
+	}
+
 	// Try to make a payment.
 	ctx.Println("Sending amount", coinValue(uint64(amount)), "to", addr)
 	err = app.Pay(addr, uint64(amount))
+
+	// Re-enable crypto wallet
+	if strings.Compare(password, "") != 0 {
```
bfbachmann: Are you doing this string comparison to check if the user is using a CryptoWallet? If so it would be cleaner to just do
```go
if app.CurrentUser.CryptoWallet {
```

Also, the `encrypt` and `decrypt` functions make sense to me, but I don't think they should be used in `send()`. If I'm a user who has enabled cryptowallet I don't want my private key on disk unencrypted. But when I go to send a transaction it will temporarily decrypt my key and save the decrypted key back to disk for a short time before restoring it to the encrypted version after the transaction is sent. I think it would be better for `send` to just decrypt and re-encrypt the wallet key in memory without touching the file.

app/user.go
```diff
@@ -35,10 +38,44 @@ func (u *User) Public() blockchain.Address {
...
+	return nil
+}
+
 // Save writes the user to a file of the given name in the current working
 // directory in JSON format. It returns an error if one occurred, or a pointer
 // to the file that was written to on success.
 func (u *User) Save(fileName string) error {
+	if _, err := os.Stat(fileName); !os.IsNotExist(err) {
+		os.Remove(fileName)
+	}
```
bfbachmann, line 77: What does this block of code do?

app/crypto.go
```diff
@@ -0,0 +1,97 @@
...
+	"strings"
+
+	"golang.org/x/crypto/pbkdf2"
+)
+
+/*
+https://golang.org/src/crypto/cipher/example_test.go
+*/
+
+const nonceSize = 12
```
bfbachmann: I prefer 
```go
const (
    nonceSize ...
)
```

app/crypto.go
```diff
@@ -0,0 +1,97 @@
...
+const saltSize = 16
+
+// Encrypt encrypts a file with a password
```
bfbachmann: I would change the word `file` here to `plainText` since you're not really dealing with a file.

app/crypto.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	cipherText := aesgcm.Seal(nil, nonce, plainText, nil)
+	cipherText = append(cipherText, salt...)
+	cipherText = append(cipherText, nonce...)
+
+	return cipherText, nil
+}
+
+// Decrypt decrypts a file with a given password
```
bfbachmann: Same here.

app/console.go
```diff
@@ -83,6 +92,92 @@ func RunConsole(a *App) *ishell.Shell {
 	return shell
 }
 
+func encrypt(ctx *ishell.Context, app *App, password string) error {
```
bfbachmann: Might want to call this function `encryptUser` or something less generic-sounding, or add a comment saying what it does.

bfbachmann (commented):
Nice job over all! I left a few suggestions.

app/console.go
bfbachmann: Same with `decrypt`.

app/console.go
bfbachmann: `secure`, `secure-wallet`, or `safe-wallet` maybe?

app/console.go
david-julien: oh man forgot about that one...

app/console.go
david-julien: trueee really good catch

app/user.go
david-julien: if the file exists already it deletes it before saving user info again, otherwise if you keep rewriting to it, the json gets all messed up

app/console.go
bfbachmann: The cleanest way to avoid this might be to just remove the `Save` calls from `encrypt` and `decrypt` and call `Save()` in `cryptoWallet()`.

david-julien: As for the ```if app.CurrentUser.CryptoWallet {```, the console needs to know if the cryptowallet was enabled and therefore decrypted before the execution of signing the tx. If it wasn't enabled before signing the transaction, or if it was enabled and then decrypted, we would get the same case where app.CurrentUser.CryptoWallet = false. Therefore if it wasn't before signing the transaction, then we would be re-encrypting for no reason.

But comparing to nil password is ugly, ill make a bool and compare to that

bfbachmann (approved):
Nice!

